PR #1, opened 2020-04-02
Add generation of hex and elf files for app, secure_fw, and bl2

oyvindronningstad:
Tested only with GNUARM, since I don't have a license set up for ARMCLANG


ioannisg (approved):
Looks good.
@oyvindronningstad do we need .hex for testing in Qemu - or only for nRF targets?

cmake/Common/CompilerArmClangCommon.cmake
```diff
@@ -186,6 +186,14 @@ function(compiler_generate_binary_output TARGET)
 	add_custom_command(TARGET ${TARGET} POST_BUILD COMMAND ${CMAKE_ARMCCLANG_FROMELF} ARGS --bincombined --output=$<TARGET_FILE_DIR:${TARGET}>/${TARGET}.bin $<TARGET_FILE:${TARGET}>)
 endfunction()
 
+function(compiler_generate_hex_output TARGET)
+	add_custom_command(TARGET ${TARGET} POST_BUILD COMMAND ${CMAKE_ARMCCLANG_FROMELF} ARGS --i32combined --output=$<TARGET_FILE_DIR:${TARGET}>/${TARGET}.hex $<TARGET_FILE:${TARGET}>)
```
ioannisg, line 190: I wonder if we could skip this, since we are going to be working with GCC only...

app/CMakeLists.txt
```diff
@@ -300,8 +300,10 @@ function(set_up_app_build)
 	endif()
 	embedded_set_target_link_includes(TARGET ${EXE_NAME} INCLUDES "${PLATFORM_LINK_INCLUDES}")
 
-	#Generate binary file from axf
+	#Generate bin/hex/elf files from axf executable
```
ioannisg: For this and _all_ similar entries: i wonder if you can just add the additional info in a new line, so the diff looks more clean.

cmake/Common/CompilerArmClangCommon.cmake
oyvindronningstad: I was thinking this could be upstreamed.

app/CMakeLists.txt
oyvindronningstad: Sure np

oyvindronningstad: Done

ioannisg (approved):
Looks ok to me
@anangl pls take a look as well

oyvindronningstad:
> do we need .hex for testing in Qemu - or only for nRF targets?

The intention was nRF targets. I haven't tried Qemu yet.

.gitignore
```diff
@@ -11,6 +11,7 @@
 
 # Garbage files
 build*/
+cmake_build*/
```
ioannisg, line 14: @oyvindronningstad just a note: this should not be up-streamed, right? I wonder if we should use commit sauces for those commits that are meant to be taken upstream

oyvindronningstad: I think it should be upstreamed. In their getting started guide it tells you to make a cmake_build dir, so it makes sense to have it in .gitignore.

I agree on the sauce tags though. Same scheme as in fw-nrfconnect-zephyr?

anangl (approved):

.gitignore
frkv: Agreed on tags like fw-nrfconnect-zephyr

frkv (approved):
Aproved pending decision on tagging (up/noup)

.gitignore
ioannisg: Let's make it simpler: just use tags for commits that need to stay down-stream only, the rest can be un-tagged

ioannisg:
> Aproved pending decision on tagging (up/noup)

Since @oyvindronningstad wants to try to upstream the .gitignore patch, I suggest we do not wait for a tag adding

oyvindronningstad:
~I will add a tag to the commits before we merge this.~ I misunderstood @ioannisg. ~Merging now.~ I'm not authorized :P, can you merge @ioannisg?

PR #2, opened 2020-04-02
Platform nrf targets support

ioannisg:
This PR is only about naming properly the nRF targets for the TF-M support for
- nRF9160
- nRF5340 APP MCU
and place the cmake switches in the existing (upstream) TF-M build configurations.

Working now on the actual CMAke targets for nrf DKs, but wanted to set in stone the target names.


ioannisg:
@anangl @oyvindronningstad pls, take a quick look

oyvindronningstad:
Did you consider making the targets chip-oriented instead of board-oriented?

ioannisg:
> Did you consider making the targets chip-oriented instead of board-oriented?

Yes I considered that, and that would make more sense for TF-M, but after talking to @carlescufi I decided to align with Zephyr, since _those_ names will be used in NCS as well.

oyvindronningstad:
> > Did you consider making the targets chip-oriented instead of board-oriented?
> 
> Yes I considered that, and that would make more sense for TF-M, but after talking to @carlescufi I decided to align with Zephyr, since _those_ names will be used in NCS as well.

Ok, I was just thinking that it makes it easier for the users if this is board-independent, so they don't have to change anything to use their own boards.

ioannisg:
> was just thinking that it makes it easier for the users if this is board-independent, so they don't have to change anything for their own

Yeap, that's correct - but for "getting people started with NCS" it is much more easy to have the same targets in TF-M and NCS, if they are testing in nRF DevKits. So it's a tradeoff and we have to go with one of the options :)

oyvindronningstad (approved):
> Yeap, that's correct - but for "getting people started with NCS" it is much more easy to have the same targets in TF-M and NCS, if they are testing in nRF DevKits. So it's a tradeoff and we have to go with one of the options :)

I agree it's simpler for getting started. FWIW If we manage to integrate TF-M as a child image into the NCS build system, I think we can make it pretty seamless :).

I can live with using the board names though.

anangl:
> > Did you consider making the targets chip-oriented instead of board-oriented?
> 
> Yes I considered that, and that would make more sense for TF-M, but after talking to @carlescufi I decided to align with Zephyr, since _those_ names will be used in NCS as well.

Perhaps then the board names could be used only in those if-s checking the value of `${TARGET_PLATFORM}`? And those could just set `PLATFORM_CMAKE_FILE` to `<cpu>.cmake`.

ioannisg:
> > > Did you consider making the targets chip-oriented instead of board-oriented?
> > 
> > 
> > Yes I considered that, and that would make more sense for TF-M, but after talking to @carlescufi I decided to align with Zephyr, since _those_ names will be used in NCS as well.
> 
> Perhaps then the board names could be used only in those if-s checking the value of `${TARGET_PLATFORM}`? And those could just set `PLATFORM_CMAKE_FILE` to `<cpu>.cmake`.

Would this lead to future problems, e.g running TF-M on nRF91 or nRF53 targets that are _not_ the DKs, and maybe have different requirements to be reflected in Cmake?
E.g. for Thingees...

anangl:
> Would this lead to future problems, e.g running TF-M on nRF91 or nRF53 targets that are _not_ the DKs, and maybe have different requirements to be reflected in Cmake?
> E.g. for Thingees...

I can't judge it at the moment. I don't know TF-M that much yet. UART pin numbers come to my mind as board specific stuff that would need to be handled in some way. But for such things we could probably use `${TARGET_PLATFORM}` again.
I don't have a strong opinion here, so we can go with what you proposed.

anangl (approved):

ioannisg:
> > Would this lead to future problems, e.g running TF-M on nRF91 or nRF53 targets that are _not_ the DKs, and maybe have different requirements to be reflected in Cmake?
> > E.g. for Thingees...
> 
> I can't judge it at the moment. I don't know TF-M that much yet. UART pin numbers come to my mind as board specific stuff that would need to be handled in some way. But for such things we could probably use `${TARGET_PLATFORM}` again.
> I don't have a strong opinion here, so we can go with what you proposed.

Agreed - we might need to revisit the names in the near future, once we have the picture of it. But I believe we will be able to change the target names without too much noise around it. Let's get this in, to continue working at the moment.

PR #3, opened 2020-04-08
platform: nrf5340: target-cfg header

ioannisg:
Commit the initial version of the target-cfg header
for nrf5340 platform. At the initial state, this
header only specifies the core platform-specific API,
not the API for peripheral security configuration.

**_Edit: duplicated this header for nRF9160._**

### DIFF
--- a/platform/ext/target/nordic_nrf/nrf5340/target_cfg.h
+++ b/platform/ext/target/nordic_nrf/nrf5340/target_cfg.h
@@ -1,0 +1,97 @@
+/*
+ * Copyright (c) 2017-2019 Arm Limited
+ * Copyright (c) 2020 Nordic Semiconductor ASA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __TARGET_CFG_H__
+#define __TARGET_CFG_H__
+
+#include "tfm_plat_defs.h"
+
+#define TFM_DRIVER_STDIO    Driver_USART0
+#define NS_DRIVER_STDIO     Driver_USART0
+
+/**
+ * \brief Store the addresses of memory regions
+ */
+struct memory_region_limits {
+    uint32_t non_secure_code_start;
+    uint32_t non_secure_partition_base;
+    uint32_t non_secure_partition_limit;
+    uint32_t veneer_base;
+    uint32_t veneer_limit;
+#ifdef BL2
+    uint32_t secondary_partition_base;
+    uint32_t secondary_partition_limit;
+#endif /* BL2 */
+};
+
+/**
+ * \brief Configures the Security Protection Unit.
+ *
+ * \return  Returns error code.
+ */
+int32_t spu_init_cfg(void);
+
+/**
+ * \brief Clears SPU interrupt.
+ */
+void spu_clear_irq(void);
+
+/**
+ * \brief Configures SAU and IDAU.
+ */
+void sau_and_idau_cfg(void);
+
+/**
+ * \brief Enables the fault handlers and sets priorities.
+ *
+ * \return Returns values as specified by the \ref tfm_plat_err_t
+ */
+enum tfm_plat_err_t enable_fault_handlers(void);
+
+/**
+ * \brief Configures the system reset request properties
+ *
+ * \return Returns values as specified by the \ref tfm_plat_err_t
+ */
+enum tfm_plat_err_t system_reset_cfg(void);
+
+/**
+ * \brief Configures the system debug properties.
+ *
+ * \return Returns values as specified by the \ref tfm_plat_err_t
+ */
+enum tfm_plat_err_t init_debug(void);
+
+/**
+ * \brief Configures all external interrupts to target the
+ *        NS state, apart for the ones associated to secure
+ *        peripherals (plus SPU)
+ *
+ * \return Returns values as specified by the \ref tfm_plat_err_t
+ */
+enum tfm_plat_err_t nvic_interrupt_target_state_cfg(void);
+
+/**
+ * \brief This function enable the interrupts associated
+ *        to the secure peripherals (plus the isolation boundary violation
+ *        interrupts)
+ *
+ * \return Returns values as specified by the \ref tfm_plat_err_t
+ */
+enum tfm_plat_err_t nvic_interrupt_enable(void);
+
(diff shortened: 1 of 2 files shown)


ioannisg:
target-cfg header contains the platform-specific, internal CORE API to do security configuration on the TrustZone-enabled core. 

For (I suppose) historical reasons, this API is _still_ implemented for each of the Cortex-M33 platforms as a non-agnostic API, even though it contains generic functionality. I suppose the reason is that the same header contains security configuration for peripherals and _that_ one is almost always vendor-specific.

Therefore, I chose to follow the same path here, and introduce the nRF5340 internal target-cfg API in the same way. Note that the initial version does not introduce the API for peripheral configuration. We can do this at a later stage. This header now only bring the minimum that is needed.

platform/ext/target/nrf5340/target_cfg.h
```diff
@@ -0,0 +1,95 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __TARGET_CFG_H__
+#define __TARGET_CFG_H__
+
+#include "platform/ext/common/uart_stdout.h"
```
anangl: Do we actually need this inclusion here?

platform/ext/target/nrf5340/target_cfg.h
```diff
@@ -0,0 +1,95 @@
...
+#include "tfm_plat_defs.h"
```
anangl: `<tfm_plat_defs.h>` would be more appropriate.

platform/ext/target/nrf5340/target_cfg.h
```diff
@@ -0,0 +1,95 @@
+/*
+ * Copyright (c) 2017-2019 Arm Limited
```
oyvindronningstad: Update copyright/add our own?

platform/ext/target/nrf5340/target_cfg.h
```diff
@@ -0,0 +1,95 @@
...
+    uint32_t secondary_partition_limit;
+#endif /* BL2 */
+};
+
+/**
+ * \brief Configures the Security Protection Unit.
+ *
+ * \return  Returns error code.
+ */
+int32_t spu_init_cfg(void);
```
anangl: I can't find that one declared/defined for other targets nor any references to it. Same story with `spu_clear_irq`. Are these indeed needed?

platform/ext/target/nrf5340/target_cfg.h
ioannisg: I copied how it is done for other platform-specific headers

platform/ext/target/nrf5340/target_cfg.h
ioannisg: Not really update - maybe add own, yes..

platform/ext/target/nrf5340/target_cfg.h
ioannisg: I think so. they are defined, for example, for AN521 platform, for the MPC, PPC controllers.
I guess it is a way to enable the interrupt and initialize the security unit. Might well be empty functions for us. I kept them for now.

platform/ext/target/nrf5340/target_cfg.h
ioannisg: I guess so - it is common inclusion for target-cfg across different ports. I kept it for this reason.

platform/ext/target/nrf5340/target_cfg.h
ioannisg: Updated :)

platform/ext/target/nrf5340/target_cfg.h
anangl: Okay, I see my mistake. I was looking for that exact name, but this function should reflect the actual name of the controller, which is other for our target than for all the others. Sorry for the noise.

ioannisg: that's fine - thanks for looking around the details

ioannisg:
@oyvindronningstad what worries me a bit is that this header should be duplicated for nrf9160 - i wonder if we should merge them to one...
RIght now each target dyplicates this target-cfg.h in the platform directory. Funny that all the MS2 targets are having almost identical copies of this header...

oyvindronningstad:
> @oyvindronningstad what worries me a bit is that this header should be duplicated for nrf9160 - i wonder if we should merge them to one...

Less copy-paste is better IMO :).

platform/ext/target/nrf5340/target_cfg.h
```diff
@@ -0,0 +1,96 @@
...
+ * limitations under the License.
+ */
+
+#ifndef __TARGET_CFG_H__
+#define __TARGET_CFG_H__
+
+#include "platform/ext/common/uart_stdout.h"
+#include "tfm_plat_defs.h"
+
+
```
anangl: You should add also definitions of `TFM_DRIVER_STDIO` and `NS_DRIVER_STDIO`, see [readme.rst](https://github.com/NordicPlayground/trusted-firmware-m/tree/master/platform/ext#target_cfgh).

ioannisg:
> > @oyvindronningstad what worries me a bit is that this header should be duplicated for nrf9160 - i wonder if we should merge them to one...
> 
> Less copy-paste is better IMO :).

I agree. For now, I committed the "framework", i.e there's a nordic_nrf parent directory, inside which there's the nrf5340 directory for the nRF53 target. For now, I want to keep the platform-cfg.h there, because I anticipate that it might not be identical for nrf5340 and nrf9160. If we see that they are identical, we can merge them and place them into a nordic_nrf/common/ folder

oyvindronningstad (approved):

platform/ext/target/nrf5340/target_cfg.h
ioannisg: Yeap, that's correct; I only wanted to do these later, once we have the names of these stdio drivers. My intention for now is to push whatever is periperhal-agnostic :)

ioannisg: @anangl is it OK to not define these macros now, since we don't know how they will be called and do it after we have defined the respective driver struct instances?

platform/ext/target/nrf5340/target_cfg.h
anangl: But maybe that inclusion was just blindly copied when new targets were added (and some old ones were used as templates)? Wouldn't it be better to add it when it appears necessary?

platform/ext/target/nrf5340/target_cfg.h
anangl: `""` means: first look in the same directory, then in include paths, `<>` means start from include paths That's why I thought the latter would be more appropriate. But that's just a detail, I won't insist on it.

platform/ext/target/nrf5340/target_cfg.h
anangl: It is OK. We could also start with both those macros just set to `Driver_USART0`.

ioannisg: OK  just brought back these defines, @anangl

anangl (approved):

PR #4, opened 2020-04-08
Fixes to make Ninja work

oyvindronningstad:
Mostly BYPRODUCTS that were missing so that Ninja was complaining it didn't know how to make the files.

The DEPFILE line seems to be copied from ARMCLANG. Removing it seems to have no ill effects.

I haven't tried running this code, just compiling.

### DIFF
--- a/cmake/Common/CompilerArmClangCommon.cmake
+++ b/cmake/Common/CompilerArmClangCommon.cmake
@@ -140,6 +140,8 @@ function(compiler_set_cmse_output TARGET FILE_PATH)
 	set_property(TARGET ${TARGET} APPEND_STRING PROPERTY LINK_FLAGS " --import_cmse_lib_out=${FILE_PATH}")
 	#Tell cmake cmse output is a generated object file.
 	SET_SOURCE_FILES_PROPERTIES("${FILE_PATH}" PROPERTIES EXTERNAL_OBJECT true GENERATED true)
+	#Dummy command to add BYPRODUCTS
+	add_custom_command(TARGET ${TARGET} POST_BUILD COMMAND "" BYPRODUCTS ${FILE_PATH})
 	#Tell cmake cmse output shall be removed by clean target.
 	get_directory_property(_ADDITIONAL_MAKE_CLEAN_FILES DIRECTORY "./" ADDITIONAL_MAKE_CLEAN_FILES)
 	set_directory_properties(PROPERTY ADDITIONAL_MAKE_CLEAN_FILES "${_ADDITIONAL_MAKE_CLEAN_FILES} ${FILE_PATH}")
--- a/cmake/Common/CompilerArmClangCommon.cmake
+++ b/cmake/Common/CompilerArmClangCommon.cmake
@@ -183,15 +183,30 @@ function(compiler_merge_library)
 endfunction()
 
 function(compiler_generate_binary_output TARGET)
-	add_custom_command(TARGET ${TARGET} POST_BUILD COMMAND ${CMAKE_ARMCCLANG_FROMELF} ARGS --bincombined --output=$<TARGET_FILE_DIR:${TARGET}>/${TARGET}.bin $<TARGET_FILE:${TARGET}>)
+	add_custom_command(TARGET ${TARGET}
+		POST_BUILD
+		COMMAND
+		${CMAKE_ARMCCLANG_FROMELF} ARGS --bincombined --output=${PROJECT_BINARY_DIR}/${TARGET}.bin $<TARGET_FILE:${TARGET}>
+		BYPRODUCTS
+		${PROJECT_BINARY_DIR}/${TARGET}.bin)
 endfunction()
 
 function(compiler_generate_hex_output TARGET)
-	add_custom_command(TARGET ${TARGET} POST_BUILD COMMAND ${CMAKE_ARMCCLANG_FROMELF} ARGS --i32combined --output=$<TARGET_FILE_DIR:${TARGET}>/${TARGET}.hex $<TARGET_FILE:${TARGET}>)
+	add_custom_command(TARGET ${TARGET}
+		POST_BUILD
+		COMMAND
+		${CMAKE_ARMCCLANG_FROMELF} ARGS --i32combined --output=${PROJECT_BINARY_DIR}/${TARGET}.hex $<TARGET_FILE:${TARGET}>
+		BYPRODUCTS
+		${PROJECT_BINARY_DIR}/${TARGET}.hex)
 endfunction()
 
 function(compiler_generate_elf_output TARGET)
-	add_custom_command(TARGET ${TARGET} POST_BUILD COMMAND ${CMAKE_ARMCCLANG_FROMELF} ARGS --elf --output=$<TARGET_FILE_DIR:${TARGET}>/${TARGET}.elf $<TARGET_FILE:${TARGET}>)
+	add_custom_command(TARGET ${TARGET}
+		POST_BUILD
+		COMMAND
+		${CMAKE_ARMCCLANG_FROMELF} ARGS --elf --output=${PROJECT_BINARY_DIR}/${TARGET}.elf $<TARGET_FILE:${TARGET}>
+		BYPRODUCTS
+		${PROJECT_BINARY_DIR}/${TARGET}.elf)
 endfunction()
 
 # Function for creating a new target that preprocesses a .c file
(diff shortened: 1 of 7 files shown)


cmake/Common/CompilerArmClangCommon.cmake
```diff
@@ -140,6 +140,8 @@ function(compiler_set_cmse_output TARGET FILE_PATH)
 	set_property(TARGET ${TARGET} APPEND_STRING PROPERTY LINK_FLAGS " --import_cmse_lib_out=${FILE_PATH}")
 	#Tell cmake cmse output is a generated object file.
 	SET_SOURCE_FILES_PROPERTIES("${FILE_PATH}" PROPERTIES EXTERNAL_OBJECT true GENERATED true)
+	#Dummy command to add BYPRODUCTS
+	add_custom_command(TARGET ${TARGET} POST_BUILD COMMAND "" BYPRODUCTS ${FILE_PATH})
```
tejlmand, line 144: Something is not right here.
The `${FILE_PATH}` already have the generated property set.

But to me, the naming indicates this is a folder and not a file, and that is something different.

Will take a closer look, but if you can provide some more information on this, it will be helpful.

tejlmand (changes requested):
Initial review looks fine, but have a small concern regarding the dummy.

Blocking to allow a little time for further investigations.

cmake/Common/CompilerArmClangCommon.cmake
oyvindronningstad: Some background:

FILE_PATH is S_VENEER_FILE

	set(VENEER_NAME ${_MY_PARAMS_VENEER_NAME}${_MY_PARAMS_POSTFIX}.o)
	...  
	set(S_VENEER_FILE "${CMAKE_CURRENT_BINARY_DIR}/${VENEER_NAME}")
	compiler_set_cmse_output(${EXE_NAME} "${S_VENEER_FILE}")

This is the only place `compiler_set_cmse_output` is called.

Ninja was failing without these lines.
Something like

    s_veneers.o missing and no rule to make it

ioannisg (approved):

cmake/Common/CompilerArmClangCommon.cmake
tejlmand: I did some more investigations on this.
The line above `SET_SOURCE_FILES_PROPERTIES("${FILE_PATH}" PROPERTIES EXTERNAL_OBJECT true GENERATED true)` was suppose to also fix this, but seems the ninja and make generator behaves slightly different.

But your suggestion should solve the same.
I did some examination of the CMake code, and the way properties are set and link dependencies makes it a bit tricky to fix in other ways, and we don't want to make to many mods to TF-M build system.

We could in theory remove the mentioned line `SET_SOURCE_FILES_PROPERTIES`, but as it does not harm, let's keep our changes to a minimum, unless we decide to send a patch upstream.

tejlmand (changes requested):
I tried building with both `make` and `ninja`.

The output size appears roughly similar (but not identical) which may be linking dependent.
`ninja` output:
```
-o bl2/ext/mcuboot/mcuboot.axf 
Memory region         Used Size  Region Size  %age Used
           FLASH:       33648 B       512 KB      6.42%
             RAM:       26432 B         2 MB      1.26%
```
`make` output:
```
-o mcuboot.axf
Memory region         Used Size  Region Size  %age Used
           FLASH:       33796 B       512 KB      6.45%
             RAM:       26432 B         2 MB      1.26%

```

but something which makes me want to do a little more investigation is the build / link order:
With `make`, at 48% build done.
```
[ 48%] Linking C executable tfm_s.axf
/opt/gcc-arm-none-eabi-7-2018-q2-update/bin/arm-none-eabi-gcc     -march=armv8-m.main  -Wl,--cmse-implib,--out-implib=/projects/github/ncs/trusted-firmware-m/build_make/secure_fw/s_veneers.o ....  -o tfm_s.axf
Memory region         Used Size  Region Size  %age Used
           FLASH:      167100 B     456896 B     36.57%
             RAM:       49372 B         1 MB      4.71%
         VENEERS:         704 B        832 B     84.62%

```
and with `ninja`, build step `227/228`
```
[227/228] : && /opt/gcc-arm-none-eabi-7-2018-q2-update/bin/arm-none-eabi-gcc  -march=armv8-m.main    -Wl,--cmse-implib,--out-implib=/projects/github/ncs/trusted-firmware-m/build_ninja/secure_fw/s_veneers.o .... -o secure_fw/tfm_s.axf 
Memory region         Used Size  Region Size  %age Used
           FLASH:      167068 B     456896 B     36.57%
             RAM:       49372 B         1 MB      4.71%
         VENEERS:         704 B        832 B     84.62%
```
It could be that this target is almost completely independent from other targets, and my current assumption is that this is the case.

I see the `s_veneers.o` is only used for linking `tfm_ns.axf ` which is done in `228/228` in ninja, and `[ 84%]` in make, which supports my assumption above.

But would like a little time to investigate further, unless you can ensure the build in both cases are correct.

cmake/Compiler/GNUARM.cmake
```diff
@@ -50,7 +50,5 @@ macro(__compiler_gnuarm lang)
     set(CMAKE_${lang}_CREATE_STATIC_LIBRARY  "<CMAKE_AR> rsc <TARGET> <LINK_FLAGS> <OBJECTS>")
     set(CMAKE_${lang}_CREATE_PREPROCESSED_SOURCE "<CMAKE_${lang}_COMPILER> <DEFINES> <INCLUDES> <FLAGS> -E <SOURCE> > <PREPROCESSED_SOURCE>")
     set(CMAKE_${lang}_CREATE_ASSEMBLY_SOURCE "<CMAKE_${lang}_COMPILER> <DEFINES> <INCLUDES> <FLAGS> -S <SOURCE> -o <ASSEMBLY_SOURCE>")
-
-    set(CMAKE_DEPFILE_FLAGS_${lang} "--depend=<DEPFILE> --depend_single_line --no_depend_system_headers")
```
tejlmand, line 54: This is a common CMake built-in feature used for toolchains.
So we should be completely sure before removing.

But one thing I do notice, is that this line is not valid for `arm-none-eabi-gcc` cause it is a `armcc` extension.
http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0472m/chr1359124912122.html

Question is if this should be changed to the corresponding GNU extension:
```
    set(CMAKE_DEPFILE_FLAGS_${lang} "-MD -MT <OBJECT> -MF <DEPFILE>")  
```
https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Compiler/GNU.cmake#L52

oyvindronningstad: I added this line, and it seems to work fine, thanks!

oyvindronningstad:
> But would like a little time to investigate further, unless you can ensure the build in both cases are correct.

Thanks very much @tejlmand for the detailed analysis.

I propose that we merge the PR, and defer investigating this to when we have working nrf platforms or until we are able to properly run another platform. I created a task for this: NCSDK-5089

oyvindronningstad:
> The output size appears roughly similar (but not identical) which may be linking dependent.
> `ninja` output:

I see the same thing:

Make:
```
[ 48%] Linking C executable tfm_s.axf
Memory region         Used Size  Region Size  %age Used
           FLASH:      166396 B     456896 B     36.42%
             RAM:       49384 B         1 MB      4.71%
         VENEERS:         704 B        832 B     84.62%
[ 48%] Built target tfm_s

...

[ 84%] Linking C executable tfm_ns.axf
Memory region         Used Size  Region Size  %age Used
           FLASH:       11168 B       447 KB      2.44%
             RAM:       13504 B         1 MB      1.29%
**[INFO]** Image version number set to 0.0.0+1
**[INFO]** Image version number set to 0.0.0+2
[ 84%] Built target tfm_ns

...

[100%] Linking C executable mcuboot.axf
Memory region         Used Size  Region Size  %age Used
           FLASH:       33972 B       512 KB      6.48%
             RAM:       26432 B         2 MB      1.26%
[100%] Built target mcuboot
```

Ninja:
```
[225/229] Linking C executable bl2/ext/mcuboot/mcuboot.axf
Memory region         Used Size  Region Size  %age Used
           FLASH:       33840 B       512 KB      6.45%
             RAM:       26432 B         2 MB      1.26%
[227/229] Linking C executable secure_fw/tfm_s.axf
Memory region         Used Size  Region Size  %age Used
           FLASH:      166364 B     456896 B     36.41%
             RAM:       49384 B         1 MB      4.71%
         VENEERS:         704 B        832 B     84.62%
[228/229] Linking C executable app/tfm_ns.axf
Memory region         Used Size  Region Size  %age Used
           FLASH:       11168 B       447 KB      2.44%
             RAM:       13504 B         1 MB      1.29%
**[INFO]** Image version number set to 0.0.0+1
**[INFO]** Image version number set to 0.0.0+2
```

It looks like Ninja defers the final linking of each image to the end.

tejlmand (approved):
Further examination so far indicates that everything should be correct.
Thus approving, and if we missed something, we just need to fix later.

oyvindronningstad:
@anangl Any opinions on this before we merge?

frkv (approved):
LGTM

anangl (approved):

oyvindronningstad:
ready to merge @ioannisg

PR #5, opened 2020-04-08
TF-M  platform: target_cfg source and SPU driver for nRF devices

ioannisg:
FYI

@anangl @oyvindronningstad I keep this draft for now

This source file implements the nRF target configuration with respect to Cortex-M33 security:
- configures interrupt security state
- configures fault handlers, priorities, reset functioning
- configures the "non-existing" SAU and SPU
- configures basic flash and ram attribution
- configures veneer region

All implementation is completed now except for the SPU configuration - for that we need to have the driver (at least the API) implemented, so we do this once, and in a clean way.

**EDIT: @anangl @oyvindronningstad this could now be reviewed**


platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,186 @@
...
+    NRF_CTRLAP->SECUREAPPROTECT.LOCK = CTRLAPPERI_SECUREAPPROTECT_LOCK_LOCK_Locked <<
+        CTRLAPPERI_SECUREAPPROTECT_LOCK_LOCK_Msk;
+
+    return TFM_PLAT_ERR_SUCCESS;
+}
+
+/*----------------- NVIC interrupt target state to NS configuration ----------*/
+enum tfm_plat_err_t nvic_interrupt_target_state_cfg(void)
+{
+    /* Target every interrupt to NS; unimplemented interrupts will be WI */
```
oyvindronningstad: "WI"? Don't know what it means :P.

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,186 @@
...
+int32_t spu_init_cfg(void)
+{
+    /* TODO
+     * Configure SPU Regions for Non-Secure Code and SRAM
+     * Configure SPU for Peripheral Security
+     * Configure Non-Secure Callable Regions
+     * Configure Secondary Image Partition for BL2
+     * Configure for non-existing SAU
+     */
+
```
oyvindronningstad, line 171: Would it be possible to copy and adapt the spm code in NCS for this function?

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,186 @@
...
+#define REGION_NAME(a, b, c) REGION(a, b, c)
+#define REGION_DECLARE(a, b, c) extern uint32_t REGION_NAME(a, b, c)
+
+/* The section names come from the scatter file */
+REGION_DECLARE(Load$$LR$$, LR_NS_PARTITION, $$Base);
+REGION_DECLARE(Load$$LR$$, LR_VENEER, $$Base);
+REGION_DECLARE(Load$$LR$$, LR_VENEER, $$Limit);
+#ifdef BL2
+REGION_DECLARE(Load$$LR$$, LR_SECONDARY_PARTITION, $$Base);
+#endif /* BL2 */
```
oyvindronningstad, line 37:
```suggestion
REGION_DECLARE(Load$$LR$$LR_VENEER$$Base);
REGION_DECLARE(Load$$LR$$LR_VENEER$$Limit);
```

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,186 @@
...
+#include "tfm_secure_api.h"
+#include "tfm_plat_defs.h"
+
+#include "log/tfm_assert.h"
+#include "log/tfm_log.h"
+
+/* Macros to pick linker symbols */
+#define REGION(a, b, c) a##b##c
+#define REGION_NAME(a, b, c) REGION(a, b, c)
+#define REGION_DECLARE(a, b, c) extern uint32_t REGION_NAME(a, b, c)
```
oyvindronningstad, line 29: It looks like this can be simplified for our use case.
```suggestion
#define REGION_DECLARE(region_name) extern uint32_t region_name
```

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,186 @@
...
+
+    .non_secure_partition_limit =
+        (uint32_t)&REGION_NAME(Load$$LR$$, LR_NS_PARTITION, $$Base) +
+        NS_PARTITION_SIZE - 1,
+
+    .veneer_base =
+        (uint32_t)&REGION_NAME(Load$$LR$$, LR_VENEER, $$Base),
+
+    .veneer_limit =
+        (uint32_t)&REGION_NAME(Load$$LR$$, LR_VENEER, $$Limit),
```
oyvindronningstad, line 55:
```suggestion
    .veneer_base =
        (uint32_t)&Load$$LR$$LR_VENEER$$Base,

    .veneer_limit =
        (uint32_t)&Load$$LR$$LR_VENEER$$Limit,
```

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,186 @@
...
+REGION_DECLARE(Load$$LR$$, LR_NS_PARTITION, $$Base);
+REGION_DECLARE(Load$$LR$$, LR_VENEER, $$Base);
+REGION_DECLARE(Load$$LR$$, LR_VENEER, $$Limit);
+#ifdef BL2
+REGION_DECLARE(Load$$LR$$, LR_SECONDARY_PARTITION, $$Base);
+#endif /* BL2 */
+
+const struct memory_region_limits memory_regions = {
+    .non_secure_code_start =
+        (uint32_t)&REGION_NAME(Load$$LR$$, LR_NS_PARTITION, $$Base) +
```
oyvindronningstad, line 41: I suggest using the defines instead, for simplicity and transparency
```suggestion
        NS_PARTITION_START +
```
This can be done for the secondary partition as well.

oyvindronningstad (changes requested):
To me it looks like the linker stuff can be simplified. Disregard my suggestions if I'm missing something :D.

ioannisg:
@oyvindronningstad I sort of completed an initial version of the target_cfg.c implementation. Please take a look again - in the mean time , I am going to respond to your early feedback :)

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: @oyvindronningstad , I have done exactly this thing! I've started with the NCS code we have for SPM, which includes TZ configuration but I defined a rather more generic API for the SPU, which allows us to use the memory_regions structure as is, and align with how other targets configure their SAU

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,203 @@
...
+{
+    /*
+     * Configure SPU Regions for Non-Secure Code and SRAM
+     * Configure SPU for Peripheral Security
+     * Configure Non-Secure Callable Regions
+     * Configure Secondary Image Partition for BL2
+     */
+
+    /* Reset Flash and SRAM configuration to all-Secure */
+    spu_regions_reset_all_secure();
```
oyvindronningstad, line 176: When is `spu_init_cfg` called? `spu_regions_reset_all_secure` seems to set the configs to their reset value, is that neccessary to do right after bootup?

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,179 @@
...
+	if (ram) {
+		/* Determine start and last ram region number */
+		size_t start_id =
+			start_addr / SRAM_SECURE_ATTRIBUTION_REGION_SIZE;
+		size_t last_id =
+			limit_addr / SRAM_SECURE_ATTRIBUTION_REGION_SIZE;
+
+		/* Configure all regions between start_id and last_id */
+		for (size_t i = start_id; i <= last_id; i++) {
+	        nrf_spu_ramregion_set(NRF_SPU, i,
```
oyvindronningstad: Fix indentation.

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,179 @@
...
+                0 /* Non-Secure */,
+                SRAM_ATTR,
+                1 /* Lock */);
+        }
+	} else {
```
oyvindronningstad: Can we decrease the duplication between the if and else?

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,179 @@
...
+	((SPU_RAMREGION_PERM_SECATTR_Secure                                    \
+	  << SPU_RAMREGION_PERM_SECATTR_Pos) &                                 \
+	 SPU_RAMREGION_PERM_SECATTR_Msk)
+
+#define SRAM_NONSEC                                                            \
+	((SPU_RAMREGION_PERM_SECATTR_Non_Secure                                \
+	  << SPU_RAMREGION_PERM_SECATTR_Pos) &                                 \
+	 SPU_RAMREGION_PERM_SECATTR_Msk)
+
+#define FLASH_ATTR (FLASH_READ | FLASH_WRITE | FLASH_EXEC)
```
oyvindronningstad: I find these names (`FLASH_ATTR`, `SRAM_ATTR`) a bit vague, i.e. I have to look them up when I encounter them in code. Suggestions: `FLASH_PERM_OPEN` `FLASH_PERM_ALL` `FLASH_RWX`

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,179 @@
...
+}
+
+/* Configure a range of (Flash or SRAM) memory regions as Non-Secure
+ *
+ * Note: region lock is applied to prevent further modification.
+ */
+void regions_config_non_secure(
+    u32_t start_addr,
+    u32_t limit_addr,
+    bool ram)
```
oyvindronningstad: Do we need this? Could check if `start_addr > 0x20000000`

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,179 @@
...
+	 * Note: Any Secure Entry functions, exposing secure services to the
+	 * Non-Secure firmware, shall be located inside this NSC area.
+	 *
+	 * If the start address of the NSC area is hard-coded, it must follow
+	 * the HW restrictions: The size must be a power of 2 between 32 and
+	 * 4096, and the end address must fall on a SPU region boundary.
+	 */
+	u32_t nsc_size = FLASH_NSC_SIZE_FROM_ADDR(start_addr);
+
+	__ASSERT(start_addr <= nsc_size,
```
oyvindronningstad:
```suggestion
	__ASSERT(size <= nsc_size,
```

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,179 @@
...
+		"The Non-Secure Callable region is overflowed by %d byte(s).\n",
+		size - nsc_size);
+
+	nrf_spu_flashnsc_set(NRF_SPU, 0,
+        FLASH_NSC_SIZE_REG(nsc_size),
+        FLASH_NSC_REGION_FROM_ADDR(start_addr),
+        1 /* Lock */);
+}
```
oyvindronningstad: Add newline

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,179 @@
...
+    for (size_t i = 0; i < NUM_SRAM_SECURE_ATTRIBUTION_REGIONS ; i++) {
+            NRF_SPU->RAMREGION[i].PERM = SRAM_ATTR | SRAM_SECURE;
+    }
+}
+
+/* Configure a range of (Flash or SRAM) memory regions as Non-Secure
+ *
+ * Note: region lock is applied to prevent further modification.
+ */
+void regions_config_non_secure(
```
oyvindronningstad: function name missing `spu_` compared to header file. Same for the next function.

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,179 @@
...
+        }
+	}
+}
+
+void regions_config_non_secure_callable_flash(u32_t start_addr, size_t size)
+{
+	/* Configure a single region in Secure Flash as Non-Secure Callable
+	 * (NSC) area.
+	 *
+	 * Area to configure is dynamically decided with help from linker code.
```
oyvindronningstad: This is not true for TF-M I think.

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,179 @@
...
+                0 /* Non-Secure */,
+                FLASH_ATTR,
+                1 /* Lock */);
+        }
+	}
+}
+
+void regions_config_non_secure_callable_flash(u32_t start_addr, size_t size)
+{
+	/* Configure a single region in Secure Flash as Non-Secure Callable
```
oyvindronningstad: Comment can be moved to header file.

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,179 @@
...
+    for (size_t i = 0; i < NUM_FLASH_SECURE_ATTRIBUTION_REGIONS ; i++) {
+            NRF_SPU->FLASHREGION[i].PERM = FLASH_ATTR | FLASH_SECURE;
+    }
+
+    for (size_t i = 0; i < NUM_SRAM_SECURE_ATTRIBUTION_REGIONS ; i++) {
+            NRF_SPU->RAMREGION[i].PERM = SRAM_ATTR | SRAM_SECURE;
+    }
+}
+
+/* Configure a range of (Flash or SRAM) memory regions as Non-Secure
```
oyvindronningstad: Comment can be moved to header file

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,203 @@
...
+
+    /* Configures SPU regions to be non-secure */
+    spu_regions_config_non_secure(memory_regions.non_secure_partition_base,
+        memory_regions.non_secure_partition_limit, 0 /* Flash */);
+    spu_regions_config_non_secure(NS_DATA_START,
+        NS_DATA_LIMIT, 1 /* SRAM */);
+
+    /* Configures veneers region to be non-secure callable */
+    spu_regions_config_non_secure_callable(memory_regions.veneer_base,
+        memory_regions.veneer_limit);
```
oyvindronningstad, line 185: The function argument should be size instead of limit.

oyvindronningstad (changes requested):
Mostly nits, overall structure looks good.

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: That's standard terminology in ARM specification: RAZ/WI: Read-as-zero/write-ignored

oyvindronningstad: Ok I see, might be worth it to write it out.

ioannisg: I would of course, have no problem to do it, although the comment is actually copy-pasted from other targets :) But it's fine to have more detailed comments, I guess

oyvindronningstad: I see, thanks :).

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: >> When is spu_init_cfg called?

This is going to be called together with `sau_and_idau_unit()` in tfm_spm_hal_init_isolation_hw()

ioannisg: > spu_regions_reset_all_secure seems to set the configs to their reset value, is that neccessary to do right after bootup?

That's a good question: so, I was thinking Tf-M should not rely on SPU register values coming out of reset, particularly when it comes to the SECURE attribution. This is just making the tf-m boot more "standalone".

oyvindronningstad: Ok, sounds good. I suggest you mention that in the comment.

ioannisg: Will do that.

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: thanks, will fix

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: Will fix - this is the problem when developing code without properlycompiling it - hopefully we get rid of this soon

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: comment added

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: Comment added now, thanks

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: No, I don't want that @oyvindronningstad 
I'll fix that in the function signature.
The reason I want the limit, is for the function to look the same as the one above, and use the corresponding memory_limit structure field as-is.

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: Yeap, this will be fixed now.

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: ASM footprint would be similar, @oyvindronningstad and now it is definitely more readable imho

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: Fixed

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: Fixed

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: fixed

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: done

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: Done

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: Done

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: The limits of the for-iteration need the if-loop to be determined. I could leave the for loop outside the if-block but then inside the for loop we would need one more if-ram conditional.

I think the duplication is somewhat acceptable and the current solution definitely has the smallest processing time

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: This is somewhat inspired and influenced by the Zephyr MPU sub-system :)
That's what we have in arm_v8m_mpu.h:

```
#define REGION_RAM_ATTR(base, size)
#define REGION_FLASH_ATTR(base, size)
```

ioannisg: But: _RWX are fine with me. I will adapt them.

ioannisg: done

platform/ext/target/nordic_nrf/Native_Driver/spu.h
```diff
@@ -0,0 +1,61 @@
...
+#include <stdint.h>
+#include <stdbool.h>
+
+/**
+ * \brief Reset all memory regions to being Secure
+ *
+ * Reset all (Flash or SRAM) memory regions to being Secure
+ * and have default (i.e. Read-Write-Execute allow) access policy
+ *
+ * Note: region lock not enforced, to allow modifying the applied configuration.
```
oyvindronningstad: Nit: This line is formatted differently for the three functions ("Note:", "enforced" vs "applied"). The note could be a dox `\note`.

I also suggest adding something like "until the next reset" for people not familiar with the locking functionality.

ioannisg: both suggestions are good, actually, I also thought of having the "until next reset cycle"
Let me apply this

ioannisg: done

ioannisg:
@oyvindronningstad I think I applied all review comments, except for the suggestions to refactor the linker section defs, which I am a bit reluctant to do now, and instead i 'd prefer to keep them as is for consistency with the rest of the targets. This makes it easier to reach a working prototype for now.

But I am not opposing any changes there - let's revisit this in a later stage of the porting, perhaps?

oyvindronningstad (approved):
Fine with me. I created task NCSDK-5136.

ioannisg:
@anangl could you also take a look here?

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,206 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "target_cfg.h"
+#include "Driver_Common.h"
+#include "platform_description.h"
+#include "device_definition.h"
```
anangl: We don't have such files yet. Are we sure we'll have them?

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,206 @@
...
+
+/*----------------- NVIC interrupt enabling for S peripherals ----------------*/
+enum tfm_plat_err_t nvic_interrupt_enable(void)
+{
+
+    /* Enable security violations outside the Cortex-M33 to trigger SPU interrupts. */
+    nrf_spu_int_enable(NRF_SPU,
+     SPU_INTENSET_FLASHACCERR_Msk |
+     SPU_INTENSET_RAMACCERR_Msk |
+     SPU_INTENCLR_PERIPHACCERR_Msk);
```
anangl: Since you are using nrfx HAL for accessing the SPU, it would be good to also use the masks defined there: `NRF_SPU_INT_*_MASK`.

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,164 @@
...
+
+#include "spu.h"
+#include <hal/nrf_spu.h>
+
+/* Platform-specific configuration */
+#define FLASH_SECURE_ATTRIBUTION_REGION_SIZE 0x8000
+#define SRAM_SECURE_ATTRIBUTION_REGION_SIZE  0x2000
+
+#define NUM_FLASH_SECURE_ATTRIBUTION_REGIONS 32
+#define NUM_SRAM_SECURE_ATTRIBUTION_REGIONS  32
```
anangl: Aren't these numbers different for nRF5340 and nRF9160 (I mean flash region size and number of RAM regions)?

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,164 @@
...
+			limit_addr / SRAM_SECURE_ATTRIBUTION_REGION_SIZE;
+
+		/* Configure all regions between start_id and last_id */
+		for (size_t i = start_id; i <= last_id; i++) {
+			nrf_spu_ramregion_set(NRF_SPU, i,
+                0 /* Non-Secure */,
+                SRAM_RWX,
+                1 /* Lock */);
+        }
+	} else {
```
anangl: Could you convert tabs to spaces, to avoid such problems in visualization of indentation?

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,164 @@
...
+#define FLASH_NSC_SIZE_FROM_ADDR(addr) FLASH_SECURE_ATTRIBUTION_REGION_SIZE    \
+	- (((u32_t)(addr)) % FLASH_SECURE_ATTRIBUTION_REGION_SIZE)
+
+#define FLASH_NSC_SIZE_REG(size) FLASH_NSC_SIZE((size) / FLASH_NSC_MIN_SIZE)
+
+
+void spu_regions_reset_all_secure(void)
+{
+    for (size_t i = 0; i < NUM_FLASH_SECURE_ATTRIBUTION_REGIONS ; i++) {
+            NRF_SPU->FLASHREGION[i].PERM = FLASH_RWX | FLASH_SECURE;
```
anangl: Why not use `nrf_spu_flashregion_set` here and `nrf_spu_ramregion_set` right below?

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,164 @@
...
+
+/* Minimum size of Non-Secure Callable regions. */
+#define FLASH_NSC_MIN_SIZE 32
+
+/* Maximum size of Non-Secure Callable regions. */
+#define FLASH_NSC_MAX_SIZE 4096
+
+
+/* Convenience macros for SPU attributes */
+#define FLASH_EXEC                                                             \
```
anangl: Why not just use `NRF_SPU_MEM_PERM_*` values?

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,164 @@
...
+	  << SPU_RAMREGION_PERM_SECATTR_Pos) &                                 \
+	 SPU_RAMREGION_PERM_SECATTR_Msk)
+
+#define FLASH_RWX (FLASH_READ | FLASH_WRITE | FLASH_EXEC)
+#define SRAM_RWX (SRAM_READ | SRAM_WRITE | SRAM_EXEC)
+
+
+#define FLASH_NSC_REGION(reg_number)                                           \
+	(((reg_number) << SPU_FLASHNSC_REGION_REGION_Pos) &                    \
+	SPU_FLASHNSC_REGION_REGION_Msk)
```
anangl: SPU HAL is responsible for doing the proper shifting and masking, so such macro should not be used in code that calls `nrf_spu_flashnsc_set()`. Same goes for `FLASH_NSC_SIZE`.

platform/ext/target/nordic_nrf/Native_Driver/spu.h
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * Configure a single region in Secure Flash as Non-Secure Callable
+ * (NSC) area.
+ *
+ * \note Any Secure Entry functions, exposing secure services to the
+ * Non-Secure firmware, shall be located inside this NSC area.
+ *
+ * If the start address of the NSC area is hard-coded, it must follow
+ * the HW restrictions: The size must be a power of 2 between 32 and
+ * 4096, and the end address must fall on a SPU region boundary.
```
anangl: Shouldn't it be checked with an assertion within the function?

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,206 @@
...
+
+    SCB->AIRCR = reg_value;
+
+    return TFM_PLAT_ERR_SUCCESS;
+}
+
+enum tfm_plat_err_t init_debug(void)
+{
+
+    volatile uint32_t *dbg_ctrl_p = (uint32_t*)DBG_CTRL_ADDR;
```
anangl: Where is `DBG_CTRL_ADDR` defined? Actually, this pointer seems to be needless, anyway.

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,206 @@
...
+#elif defined(DAUTH_NS_ONLY)
+    /* Allow debugging Non-Secure only */
+    NRF_CTRLAP->APPROTECT.DISABLE = CTRLAPPERI_APPROTECT_DISABLE_KEY_Msk;
+    NRF_CTRLAP->SECUREAPPROTECT.DISABLE = 0;
+#elif defined(DAUTH_FULL)
+    /* Allow debugging */
+    NRF_CTRLAP->APPROTECT.DISABLE = CTRLAPPERI_APPROTECT_DISABLE_KEY_Msk;
+    NRF_CTRLAP->SECUREAPPROTECT.DISABLE = CTRLAPPERI_SECUREAPPROTECT_DISABLE_KEY_Msk;
+#else
+#if !defined(DAUTH_CHIP_DEFAULT)
```
anangl:
```suggestion
#elif !defined(DAUTH_CHIP_DEFAULT)
```

anangl (changes requested):

platform/ext/target/nordic_nrf/Native_Driver/spu.c
oyvindronningstad: Yes.
Flash region size is 32 for nrf5340pdk and 16 for nrf5340dk (Erratum 19)
Number of RAM regions is double for nrf5340 vs nrf9160 since RAM size is double.

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,206 @@
+/*
```
oyvindronningstad, line 1: Did you consider having this file be common for 91 and 53? If it can't be, will you add one for 91 here?

platform/ext/target/nordic_nrf/Native_Driver/spu.c
```diff
@@ -0,0 +1,164 @@
+/*
```
oyvindronningstad: Can we move `Native_Driver` into `common`?

    nordic_nrf/common/Native_Driver

I think it makes more sense to keep everything that is common in the `common` directory.

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: I guess we do not care about the _dk_ with the Erratum19 @oyvindronningstad , can we assume we only care about the nRF5340pdk?

oyvindronningstad: It's the pdk that has the erratum (16 is correct, 32 is faulty). We definitely need to care about the dk when it comes.

ioannisg: Anyways, I also thought about it during development, and had decided to ignore this for now, but since you both bring this up, maybe let's fix this.
For that we need #ifdef 's here, though. @oyvindronningstad which are or shall be the directives here?

ioannisg: btw, i am also thinking these should be in some internal header, not in spu.c, any ideas?

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: the reason is "for making the code more compressed". nrf_spu_flashregion_set would need to be called for each region.
a similar practice is followed (and for all tf-m targets) in `nvic_interrupt_target_state_cfg`

platform/ext/target/nordic_nrf/Native_Driver/spu.h
ioannisg: Yes, this is actually done in the downstream SPM - I was just thinking to skip for the first prototype, but I can add the assert (need to find out how TF-M asserts work)

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: I really don't want too many common stuff at the moment, considering that platform code duplication is an accepted practice and this file is _anyways_ platform-specific (even though in its current state it still does not seem so)

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: That for @anangl to decide, actually.

platform/ext/target/nordic_nrf/Native_Driver/spu.c
anangl: I'm not sure I understand what you mean. `nrf_spu_flashregion_set` is an inline function, so the generated code will be the same (unless assertions are enabled).

platform/ext/target/nordic_nrf/Native_Driver/spu.c
anangl: Yes, let's place it in the common directory. I'd also suggest to name it `native_drivers` (we'll have there also `cmsis_drivers`).

platform/ext/target/nordic_nrf/Native_Driver/spu.c
anangl: We've got `SPU_FLASH_REGION_SIZE` in `region_defs.h`. We could add a corresponding definition for RAM regions. And number of regions can be calculated from the total size of the respective memory, can't it?

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: Sure, done now.

ioannisg: Actually, I moved the implementation inside the SPU driver, and I kept a wrapper here only. This looks more how it is done for other targets, and will allow us later to more easily convert the SPU API into a standard "driver" API..

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: @anangl  `nrf_spu_flashregion_set` is inline, but contains two conditionals in the assignment operation (for secure attribute and lock booleans) which in certain cases would need to be evaluated in runtime, increasing the number of instructions executed. Most likely when passing constants, the compiler would optimize-out the '?' operators but this is not guaranteed by C code.

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: done now, thanks.

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: Sure, I just did not know these macros exist. Will use them, thanks @anangl

ioannisg: This, together with the comment above, @anangl @oyvindronningstad allows me to get rid of all locally-defined convenience macros for SPU permissions

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: Finally, fixed according to your proposal, as it allows me to get rid of the locally-defined macros for attributes

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: Agreed, that should be fixed.
@oyvindronningstad ^^

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: @oyvindronningstad ^^

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: @anangl btw,
```
    p_reg->FLASHNSC[flash_nsc_id].REGION = (uint32_t)region_number |
        (lock_conf ? SPU_FLASHNSC_REGION_LOCK_Msk : 0);
    p_reg->FLASHNSC[flash_nsc_id].SIZE = (uint32_t)flash_nsc_size |
        (lock_conf ? SPU_FLASHNSC_SIZE_LOCK_Msk : 0);
```

there is no shift, mask applied on region and size, but i guess this is OK if shift should be 0. But strictly speaking, the shifting here is (was) not redundant.

ioannisg: Fixed now.

platform/ext/target/nordic_nrf/Native_Driver/spu.h
ioannisg: EDIT: these asserts are not part of downstream (NCS SPM) run-time checking. So I added the checks as NRF_ASSERT statements. [Checks are done at build time during linking, though]

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: > We've got `SPU_FLASH_REGION_SIZE` in `region_defs.h`. We could add a corresponding definition for RAM regions. And number of regions can be calculated from the total size of the respective memory, can't it?

If you know the region size, yes.

ioannisg: Alright, I am adding the SPU_SRAM_REGION_SIZE for now in region_defs.h.
I am also calculating the number of regions as proposed.

Let's not bother, yet, about _pdk and _dk @oyvindronningstad , this should come once we have different cmake targets for these, or until we have a way to detect the DK version in runtime.

For the first prototype this should not be a blocker imho.

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: Might, yes. At least device_definition.h is used to pass buld-time configuration, which in Zephyr is passed via Kconfig, DTS.

ioannisg: @anangl would you prefer adding these headers on demand?

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: Yeap, that 's copy-paste error. Removed now.

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: applied, thanks

platform/ext/target/nordic_nrf/Native_Driver/spu.c
ioannisg: Sure. I am doing this on a separate commit to track the move.

ioannisg: EDIT - done now (on a single commit)

ioannisg: side note @anangl it's a bit messy across the different targets, regarding the nomenclature around cmsis and native driver directory names. Several targets use Native_Driver and CMSIS_Driver, while other use cmsis_drivers and native_drivers.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
```diff
@@ -0,0 +1,130 @@
...
+    u32_t start_addr,
+    u32_t limit_addr,
+    bool ram)
+{
+	if (ram) {
+		/* Determine start and last ram region number */
+		size_t start_id =
+			start_addr / SRAM_SECURE_ATTRIBUTION_REGION_SIZE;
+		size_t last_id =
+			limit_addr / SRAM_SECURE_ATTRIBUTION_REGION_SIZE;
```
anangl: It won't work like that. You need to subtract the RAM base address from both these addresses.

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Reset all (Flash or SRAM) memory regions to being Secure
+ * and have default (i.e. Read-Write-Execute allow) access policy
+ *
+ * \note region lock is not applied to allow modifying the configuration.
+ */
+void spu_regions_reset_all_secure(void);
+
+/**
+ * \brief Configre memory regions as Non-Secure
```
anangl: Configure

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Configure a range of (Flash or SRAM) memory regions as Non-Secure
+ *
+ * \note region lock is applied to prevent further modification during
+ * the current reset cycle.
+ */
+
+void spu_regions_config_non_secure( u32_t start_addr, u32_t limit_addr, bool ram);
```
anangl: This API is a bit inconsistent. We've got a common function for configuring non-secure regions in flash and RAM but separate functions for configuring NSC regions (the one for RAM non-existing yet).
There is actually no advantage of having this `ram` parameter here as there is no common code for flash and RAM in this function. Thus, I think it would be better to have separate functions for configuring non-secure flash and RAM sections. Additionally, if one of them was not used in some fancy setup, the other one could be discarded by the linker.

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,197 @@
...
+#include "target_cfg.h"
+#include "Driver_Common.h"
+#include "platform_description.h"
+#include "device_definition.h"
+#include "region_defs.h"
+#include "tfm_secure_api.h"
+#include "tfm_plat_defs.h"
+
+#include "log/tfm_assert.h"
+#include "log/tfm_log.h"
```
anangl: Those two do not seem to be needed here.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
```diff
@@ -0,0 +1,130 @@
...
+	(TOTAL_RAM_SIZE / SRAM_SECURE_ATTRIBUTION_REGION_SIZE)
+
+/* Convenience macros for SPU NCS attribution */
+#define FLASH_NSC_REGION_FROM_ADDR(addr) \
+	((u32_t)addr / FLASH_SECURE_ATTRIBUTION_REGION_SIZE)
+
+#define FLASH_NSC_SIZE_FROM_ADDR(addr) (FLASH_SECURE_ATTRIBUTION_REGION_SIZE \
+	- (((u32_t)(addr)) % FLASH_SECURE_ATTRIBUTION_REGION_SIZE))
+
+#define FLASH_NSC_SIZE_REG(size) ((31 - __builtin_clz(size)) - 4)
```
anangl, line 55: Some comment explaining briefly this implementation would be nice here.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
```diff
@@ -0,0 +1,130 @@
...
+	/* Check Non-Secure Callable region possible overflow */
+	NRFX_ASSERT(size <= nsc_size);
+
+	/* Check Non-Secure Callable region ending on SPU boundary */
+	NRFX_ASSERT(((start_addr + size) %
+			FLASH_SECURE_ATTRIBUTION_REGION_SIZE) == 0);
+
+	/* Check Non-Secure Callable region power-of-2 size compliance */
+	NRFX_ASSERT((nsc_size & (nsc_size - 1)) == 0,
+		"The Non-Secure Callable region size not power-of-2\n");
```
anangl: `NRFX_ASSERT` takes only one parameter.

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
anangl: Definitely. Otherwise, this file will be non-compilable until these included files are added to the tree.

ioannisg: Alright. Removed the includes now (may add later)

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
ioannisg: Thanks, fixed.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
ioannisg: Yeap, that was forgotten; corrected it now

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
ioannisg: > Additionally, if one of them was not used in some fancy setup, the other one could be discarded by the linker.

This is very unlikely @anangl , I cannot think of us doing a non XIP Zephyr build on nRF platforms (that would be the only reason not to configure non-secure flash, and even this would not work if we wanted to do non-secure image updades e.g. via MCUboot...)

ioannisg: > Thus, I think it would be better to have separate functions for configuring non-secure flash and RAM sections.

I will try that, if this makes the API consistent, in your opinion. The code overhead should not be a lot. (But there will be some.)

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
ioannisg: Comments added now.
The description was minimal, following what we had in NCS, but agree, there's always room for improvement :)

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: I guess, same as for any other inclusions - let's add them on an on-demand basis
Removed now.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
ioannisg: Oh, that's good catch, @anangl thanks!
Fixing this now

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
oyvindronningstad: Ok that's fine, but FWIW I thought it might be a little easier with one common file instead of two during development.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
```diff
@@ -0,0 +1,143 @@
...
+{
+	size_t size = limit_addr - start_addr + 1;
+
+	u32_t nsc_size = FLASH_NSC_SIZE_FROM_ADDR(start_addr);
+
+	/* Check Non-Secure Callable region possible overflow */
+	NRFX_ASSERT(size <= nsc_size);
+
+	/* Check Non-Secure Callable region ending on SPU boundary */
+	NRFX_ASSERT(((start_addr + size) %
```
oyvindronningstad: I think you want 
```suggestion
	NRFX_ASSERT(((start_addr + nsc_size) %
```

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
```diff
@@ -0,0 +1,143 @@
...
+			FLASH_SECURE_ATTRIBUTION_REGION_SIZE) == 0);
+
+	/* Check Non-Secure Callable region power-of-2 size compliance */
+	NRFX_ASSERT((nsc_size & (nsc_size - 1)) == 0);
```
oyvindronningstad, line 144: Since we are doing comprehensive checking, we could also check that nsc_size is between 32 and 4096.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
```diff
@@ -0,0 +1,143 @@
...
+ * Determine the NSC region size based on a given NCS region base address.
+ */
+#define FLASH_NSC_SIZE_FROM_ADDR(addr) (FLASH_SECURE_ATTRIBUTION_REGION_SIZE \
+	- (((u32_t)(addr)) % FLASH_SECURE_ATTRIBUTION_REGION_SIZE))
+
+/*
+ * Determine the encoded the SPU NCS Region Size value,
+ * based on the absolute NCS region size in bytes.
+ */
+#define FLASH_NSC_SIZE_REG(size) ((31 - __builtin_clz(size)) - 4)
```
oyvindronningstad, line 55: I think this gives the wrong value if size is a power of 2 and if it is less than 16.
We want
`FLASH_NSC_SIZE_REG(0) == FLASH_NSC_SIZE_REG(15) == FLASH_NSC_SIZE_REG(31) == FLASH_NSC_SIZE_REG(32) == 1`
and
`FLASH_NSC_SIZE_REG(33) == FLASH_NSC_SIZE_REG(64) == 2` etc.
https://infocenter.nordicsemi.com/topic/ps_nrf9160/spu.html?cp=2_0_0_5_14_8_16#register.FLASHNSC-0-1.SIZE
```suggestion
#define FLASH_NSC_SIZE_REG(size) (33 - __builtin_clz((size - 1)>>5))
```
`size - 1` fixes for powers of 2, while the other changes fix for less than 16.
We should probably refer to the PS in the comment otherwise this macro will be confusing for readers :P.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
ioannisg: Fixed

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
ioannisg: Updated the API, to have separate functions for flash and sram NS configuration

ioannisg (commented):
@anangl the PR is now revised based on your latest feedback

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,191 @@
...
+    spu_regions_sram_config_non_secure(NS_DATA_START, NS_DATA_LIMIT);
+
+    /* Configures veneers region to be non-secure callable */
+    spu_regions_flash_config_non_secure_callable(memory_regions.veneer_base,
+        memory_regions.veneer_limit);
+
+#ifdef BL2
+    /* Secondary image partition */
+    spu_regions_flash_config_non_secure(memory_regions.secondary_partition_base,
+        memory_regions.secondary_partition_limit, 0 /* Flash */);
```
anangl: Remove the last parameter.

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
```diff
@@ -0,0 +1,80 @@
...
+#define __SPU_H__
+
+#include <stddef.h>
+#include <stdint.h>
+#include <stdbool.h>
+
+/**
+ * \brief SPU interrupt enabling
+ *
+ * Enable security violations outside the Cortex-M33
```
oyvindronningstad, line 27: "outside the Cortex-33" means in the peripherals?

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * \brief Configure Non-Secure Callable area
+ *
+ * Configure a single region in Secure Flash as Non-Secure Callable
+ * (NSC) area.
+ *
+ * \note Any Secure Entry functions, exposing secure services to the
+ * Non-Secure firmware, shall be located inside this NSC area.
+ *
+ * If the start address of the NSC area is hard-coded, it must follow
```
oyvindronningstad, line 71: This function does this rounding up itself, right? So the restrictions on start_addr and limit_addr are:

Must be within a 4096 byte area at the end of an SPU region

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,191 @@
...
+        NVIC->ITNS[i] = 0xFFFFFFFF;
+    }
+
+    /* Make sure that the SPU is targeted to S state */
+    NVIC_ClearTargetState(SPU_IRQn);
+
+    return TFM_PLAT_ERR_SUCCESS;
+}
+
+/*----------------- NVIC interrupt enabling for S peripherals ----------------*/
```
oyvindronningstad, line 140: Which 5 peripherals?

oyvindronningstad (changes requested):

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * \brief Configure Flash memory regions as Non-Secure
+ *
+ * Configure a range of Flash memory regions as Non-Secure
+ *
+ * \note region lock is applied to prevent further modification during
+ * the current reset cycle.
+ */
+void spu_regions_flash_config_non_secure( u32_t start_addr, u32_t limit_addr);
```
anangl: `s/u32_t/uint32_t/`

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,191 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "target_cfg.h"
+#include "Driver_Common.h"
+#include "region_defs.h"
+#include "tfm_secure_api.h"
+#include "tfm_plat_defs.h"
```
anangl, line 21: Add `#include <spu.h>`

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -0,0 +1,191 @@
...
+    SCB->AIRCR = reg_value;
+
+    return TFM_PLAT_ERR_SUCCESS;
+}
+
+enum tfm_plat_err_t init_debug(void)
+{
+#if defined(DAUTH_NONE)
+    /* Disable debugging */
+    NRF_CTRLAP->APPROTECT.DISABLE = 0;
```
anangl, line 104: You need to either use `NRF_CTRLAP_S` or include `<nrfx.h>` to get this symbol translated.

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: Done, thanks.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
ioannisg: @oyvindronningstad @anangl may we, instead, agree that this is the responsibility of 
- the linker
- the developer (i.e. to provide a valid size)?

This is how it is done in SPM as well. Also, this is how the function is documented:
```
 If the start address of the NSC area is hard-coded, it must follow
 * the HW restrictions: The size must be a power of 2 between 32 and
 * 4096, and the end address must fall on a SPU region boundary.
```
I think all the above  should be checked with proper ASSERTS. I added an ASSERT for the range checks. So now all the "restrictions" are checked in run-time.

I also enhanced the description of the macro:
```
/*
 * Determine the encoded the SPU NCS Region Size value,
 * based on the absolute NCS region size in bytes.
 *
 * size shall be a valid SPU NCS Region size value
 */
```
so the documentation now warns that we need to pass a valid size.

I don't want this macro to do truncation either left or right, I want this macro to only do the encoding of the size value.

#define FLASH_NSC_SIZE_REG(size) ((31 - __builtin_clz(size)) - 4)

So I think we should keep the define as is, and solve our size-checking via ASSERTS, Documentation and Linker.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
ioannisg: Yes, done!

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
ioannisg: Yes. also DMA

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
ioannisg: Nope, the function does no rounding. This is linker's responsibility OR, the responsibility of the flash layout (based on hard-coded defines or cmake)

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
ioannisg: yes, done already

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: That's an "S" I think this sentence was copy-pasted from other targets

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: Isn't this going to be done via Driver_Common.h?

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: I guess, I would include nrfx.h - can we expect this is done in Driver_Common.h? 
I can add this inclusion, anyway.

ioannisg: added

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
ioannisg: Done, and in all other places, as well. I guess it's hard with PR revision without proper compilation/linking...

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
anangl: I don't think so. It's a common file to be included by all CMSIS driver implementations and it's not related to SPU. BTW its inclusion does not seem to be needed here.

ioannisg: Alright, so I'll replace DriverCommon.h inclusion with spu.h

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
oyvindronningstad: Haha, right 😅

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: done

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
```diff
@@ -0,0 +1,153 @@
...
+#define FLASH_SECURE_ATTRIBUTION_REGION_SIZE SPU_FLASH_REGION_SIZE
+#define SRAM_SECURE_ATTRIBUTION_REGION_SIZE  SPU_SRAM_REGION_SIZE
+
+#define NUM_FLASH_SECURE_ATTRIBUTION_REGIONS \
+	(FLASH_TOTAL_SIZE / FLASH_SECURE_ATTRIBUTION_REGION_SIZE)
+#define NUM_SRAM_SECURE_ATTRIBUTION_REGIONS \
+	(TOTAL_RAM_SIZE / SRAM_SECURE_ATTRIBUTION_REGION_SIZE)
+
+#define DEVICE_FLASH_BASE_ADDRESS FLASH_BASE_ADDRESS
+#define DEVICE_SRAM_BASE_ADDRESS SRAM_BASE_ADDRESS
```
anangl, line 31: `RAM_BASE_ADDRESS`

anangl (approved):
LGTM now, modulo this typo pointed out above.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
ioannisg: Oh right. Fixed now. Actually I kept the "SRAM" prefix, since this is used allover the place.

ioannisg: So I had to check the macros defined in flash_layout.h

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
oyvindronningstad: Yep, sorry, I forgot that the parameter has been checked before calling this macro.

I guess we should provide the linker changes to tfm_common_s.ld.template, will you do that in this PR?

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
oyvindronningstad: Will you add an nRF9160 variant in this PR?

oyvindronningstad (approved):
It's time to get this in IMO :). But if you want to build it before we merge it, let's merge #12 first.

PR #6, opened 2020-04-14
platform: nrf5340: adding flash layout information

ioannisg:
We are introducing the flash layout information for the
nRF5340 target, to be used in memory partitioning.

- Pushing this for feedback, while working on region_defs.h header.
flash_layout.h should introduce the flash partitioning for nRF530 APP MCU target.
- The default partitioning is meant to work with upstream Zephyr samples/tests.
- Care has been taken so this partitioning looks like upstream Zephyr flash partitioning, so TF-M is compatible with Zephyr. (But note that, at the end of the day, any upstream overlays, and/or partition manager (downstream) may override flash partitioning.)

- Secure Storage and Internal Trusted Firmware partitions are placed on the flash area that Zephyr defines as image scratch.

- **_update_**: pushed the region_defs.h. as well.

### DIFF
--- a/platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
+++ b/platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
@@ -1,0 +1,238 @@
+/*
+ * Copyright (c) 2018-2020 Arm Limited. All rights reserved.
+ * Copyright (c) 2020 Nordic Semiconductor ASA. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __FLASH_LAYOUT_H__
+#define __FLASH_LAYOUT_H__
+
+/* Flash layout on NRF5340 Application MCU with BL2:
+ *
+ * 0x0000_0000 BL2 - MCUBoot (64 KB)
+ * 0x0001_0000 Primary image area (448 KB):
+ *    0x0001_0000 Secure     image primary (192 KB)
+ *    0x0004_0000 Non-secure image primary (256 KB)
+ * 0x0008_0000 Secondary image area (448 KB):
+ *    0x0008_0000 Secure     image secondary (192 KB)
+ *    0x000b_0000 Non-secure image secondary (256 KB)
+ * 0x000f_0000 Secure Storage Area (16 KB)
+ * 0x000f_4000 Internal Trusted Storage Area (8 KB)
+ * 0x000f_6000 NV counters area (4 KB)
+ * 0x000f_7000 Unused
+ *
+ * Flash layout on NRF5340 Application MCU without BL2:
+ *
+ * 0x0000_0000 Primary image area (960 KB):
+ *    0x0000_0000 Secure     image primary (480 KB)
+ *    0x0007_8000 Non-secure image primary (480 KB)
+ * 0x000f_0000 Secure Storage Area (16 KB)
+ * 0x000f_4000 Internal Trusted Storage Area (8 KB)
+ * 0x000f_6000 NV counters area (4 KB)
+ * 0x000f_7000 Unused
+ */
+
+/* This header file is included from linker scatter file as well, where only a
+ * limited C constructs are allowed. Therefore it is not possible to include
+ * here the platform_base_address.h to access flash related defines. To resolve
+ * this some of the values are redefined here with different names, these are
+ * marked with comment.
+ */
+
+/* Size of a Secure and of a Non-secure image */
+#define FLASH_S_PARTITION_SIZE                (0x30000)       /* S partition: 192 kB*/
+#define FLASH_NS_PARTITION_SIZE               (0x40000)       /* NS partition: 256 kB*/
+#define FLASH_MAX_PARTITION_SIZE        ((FLASH_S_PARTITION_SIZE >   \
+                                          FLASH_NS_PARTITION_SIZE) ? \
+                                         FLASH_S_PARTITION_SIZE :    \
+                                         FLASH_NS_PARTITION_SIZE)
+
+/* Sector size of the embedded flash hardware (erase/program) */
+#define FLASH_AREA_IMAGE_SECTOR_SIZE        (0x1000)           /* 4 KB. Flash memory program/erase operations have a page granularity. */
+
+/* FLASH size */
+#define FLASH_TOTAL_SIZE                    (0x100000)    /* 1024 kB. */
+
+/* Flash layout info for BL2 bootloader */
+#define FLASH_BASE_ADDRESS                  (0x00000000)
+
+#ifdef BL2
+/* Offset and size definitions of the flash partitions that are handled by the
+ * bootloader. The image swapping is done between IMAGE_PRIMARY and
+ * IMAGE_SECONDARY, SCRATCH is used as a temporary storage during image
+ * swapping.
+ */
+#define FLASH_AREA_BL2_OFFSET      (0x0)
+#define FLASH_AREA_BL2_SIZE        (0x10000) /* 64 KB */
+
+#if !defined(MCUBOOT_IMAGE_NUMBER) || (MCUBOOT_IMAGE_NUMBER == 1)
+/* Secure + Non-secure image primary slot */
+#define FLASH_AREA_0_ID            (1)
+#define FLASH_AREA_0_OFFSET        (FLASH_AREA_BL2_OFFSET + FLASH_AREA_BL2_SIZE)
+#define FLASH_AREA_0_SIZE          (FLASH_S_PARTITION_SIZE + \
+                                    FLASH_NS_PARTITION_SIZE)
+/* Secure + Non-secure secondary slot */
+#define FLASH_AREA_2_ID            (FLASH_AREA_0_ID + 1)
+#define FLASH_AREA_2_OFFSET        (FLASH_AREA_0_OFFSET + FLASH_AREA_0_SIZE)
+#define FLASH_AREA_2_SIZE          (FLASH_S_PARTITION_SIZE + \
+                                    FLASH_NS_PARTITION_SIZE)
+/* Not used, only the Non-swapping firmware upgrade operation
+ * is supported on NRF5340 Application MCU.
+ */
+#define FLASH_AREA_SCRATCH_ID      (FLASH_AREA_2_ID + 1)
+#define FLASH_AREA_SCRATCH_OFFSET  (FLASH_AREA_2_OFFSET + FLASH_AREA_2_SIZE)
+#define FLASH_AREA_SCRATCH_SIZE    (0)
+/* Maximum number of image sectors supported by the bootloader. */
+#define MCUBOOT_MAX_IMG_SECTORS    ((FLASH_S_PARTITION_SIZE + \
+                                     FLASH_NS_PARTITION_SIZE) / \
+                                    FLASH_AREA_IMAGE_SECTOR_SIZE)
+#elif (MCUBOOT_IMAGE_NUMBER == 2)
+/* Secure image primary slot */
+#define FLASH_AREA_0_ID            (1)
+#define FLASH_AREA_0_OFFSET        (FLASH_AREA_BL2_OFFSET + FLASH_AREA_BL2_SIZE)
+#define FLASH_AREA_0_SIZE          (FLASH_S_PARTITION_SIZE)
+/* Non-secure image primary slot */
+#define FLASH_AREA_1_ID            (FLASH_AREA_0_ID + 1)
+#define FLASH_AREA_1_OFFSET        (FLASH_AREA_0_OFFSET + FLASH_AREA_0_SIZE)
+#define FLASH_AREA_1_SIZE          (FLASH_NS_PARTITION_SIZE)
+/* Secure image secondary slot */
+#define FLASH_AREA_2_ID            (FLASH_AREA_1_ID + 1)
+#define FLASH_AREA_2_OFFSET        (FLASH_AREA_1_OFFSET + FLASH_AREA_1_SIZE)
+#define FLASH_AREA_2_SIZE          (FLASH_S_PARTITION_SIZE)
+/* Non-secure image secondary slot */
+#define FLASH_AREA_3_ID            (FLASH_AREA_2_ID + 1)
+#define FLASH_AREA_3_OFFSET        (FLASH_AREA_2_OFFSET + FLASH_AREA_2_SIZE)
+#define FLASH_AREA_3_SIZE          (FLASH_NS_PARTITION_SIZE)
+/* Not used, only the Non-swapping firmware upgrade operation
+ * is supported on NRF5340 Application MCU.
+ */
+#define FLASH_AREA_SCRATCH_ID      (FLASH_AREA_3_ID + 1)
+#define FLASH_AREA_SCRATCH_OFFSET  (FLASH_AREA_3_OFFSET + FLASH_AREA_3_SIZE)
+#define FLASH_AREA_SCRATCH_SIZE    (0)
+/* Maximum number of image sectors supported by the bootloader. */
+#define MCUBOOT_MAX_IMG_SECTORS    (FLASH_MAX_PARTITION_SIZE / \
+                                    FLASH_AREA_IMAGE_SECTOR_SIZE)
+#else /* MCUBOOT_IMAGE_NUMBER > 2 */
+#error "Only MCUBOOT_IMAGE_NUMBER 1 and 2 are supported!"
+#endif /* MCUBOOT_IMAGE_NUMBER */
+
+/* Not used, only the Non-swapping firmware upgrade operation
+ * is supported on nRF5340. The maximum number of status entries
+ * supported by the bootloader.
+ */
+#define MCUBOOT_STATUS_MAX_ENTRIES      (0)
+
+#else
+
+/* Secure + Non-secure image primary slot */
+#define FLASH_AREA_0_ID            (1)
+#define FLASH_AREA_0_OFFSET        (0x0)
+#define FLASH_AREA_0_SIZE          (FLASH_S_PARTITION_SIZE + \
+                                    FLASH_NS_PARTITION_SIZE)
+
+#endif /* BL2 */
+
+#define FLASH_SST_AREA_OFFSET           (FLASH_AREA_SCRATCH_OFFSET + \
+                                         FLASH_AREA_SCRATCH_SIZE)
+#define FLASH_SST_AREA_SIZE             (0x4000)   /* 16 KB */
+
+/* Internal Trusted Storage (ITS) Service definitions */
+#define FLASH_ITS_AREA_OFFSET           (FLASH_SST_AREA_OFFSET + \
+                                         FLASH_SST_AREA_SIZE)
+#define FLASH_ITS_AREA_SIZE             (0x2000)   /* 8 KB */
+
+/* NV Counters definitions */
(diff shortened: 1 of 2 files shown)


platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
```diff
@@ -0,0 +1,242 @@
...
+ * 0x000f_6200 Unused
+ *
+ * Flash layout on NRF5340 Application MCU without BL2:
+ *
+ * 0x0000_0000 Primary image area (960 KB):
+ *    0x0000_0000 Secure     image primary (480 KB)
+ *    0x0007_8000 Non-secure image primary (480 KB)
+ * 0x000f_0000 Secure Storage Area (16 KB)
+ * 0x000f_4000 Internal Trusted Storage Area (8 KB)
+ * 0x000f_6000 NV counters area (4 KB)
```
oyvindronningstad, line 42: Why the difference in size with vs without BL2? (512 B vs 4 KB)

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
```diff
@@ -0,0 +1,160 @@
...
+#define BL2_MSP_STACK_SIZE      (0x0001800)
+
+#define S_HEAP_SIZE             (0x0001000)
+#define S_MSP_STACK_SIZE_INIT   (0x0000400)
+#define S_MSP_STACK_SIZE        (0x0000800)
+#define S_PSP_STACK_SIZE        (0x0000800)
+
+#define NS_HEAP_SIZE            (0x0001000)
+#define NS_MSP_STACK_SIZE       (0x0000400)
+#define NS_PSP_STACK_SIZE       (0x0000C00)
```
anangl: Was this intentional to use 7 hex digits here? I think it's better to either use 8 (or 4) digits or omit the leading zeros, to avoid potential confusion.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
```diff
@@ -0,0 +1,160 @@
...
+#define NS_CODE_LIMIT   (NS_CODE_START + NS_CODE_SIZE - 1)
+
+#define NS_DATA_START   (NS_RAM_ALIAS(TOTAL_RAM_SIZE / 2))
+#define NS_DATA_SIZE    (TOTAL_RAM_SIZE / 2)
+#define NS_DATA_LIMIT   (NS_DATA_START + NS_DATA_SIZE - 1)
+
+/* NS partition information is used for SPU configuration */
+#define NS_PARTITION_START \
+            (NS_ROM_ALIAS(NS_IMAGE_PRIMARY_PARTITION_OFFSET))
+
```
anangl: Probably unwanted empty line.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
```diff
@@ -0,0 +1,160 @@
...
+                                           + FLASH_S_PARTITION_SIZE)
+#else
+#define NS_IMAGE_PRIMARY_PARTITION_OFFSET (FLASH_AREA_2_OFFSET \
+                                           + FLASH_S_PARTITION_SIZE)
+#endif /* !LINK_TO_SECONDARY_PARTITION */
+
+/* Boot partition structure if MCUBoot is used:
+ * 0x0_0000 Bootloader header
+ * 0x0_0400 Image area
+ * 0x1_FC00 Trailer
```
anangl: I guess it should be `0x0_FC00`, as the [MCUBoot area size](https://github.com/NordicPlayground/trusted-firmware-m/pull/6/files#diff-1f387a5563b0b082bfcd63e10f69d8e6R77) is `(0x10000) /* 64 KB */`.

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
```diff
@@ -0,0 +1,242 @@
...
+
+/* In this target the CMSIS driver requires only the offset from the base
+ * address instead of the full memory address.
+ */
+#define SST_FLASH_AREA_ADDR     FLASH_SST_AREA_OFFSET
+/* Dedicated flash area for SST */
+#define SST_FLASH_AREA_SIZE     FLASH_SST_AREA_SIZE
+#define SST_SECTOR_SIZE         FLASH_AREA_IMAGE_SECTOR_SIZE
+/* Number of SST_SECTOR_SIZE per block */
+#define SST_SECTORS_PER_BLOCK   (0x8)
```
anangl: This would mean that the block size is `0x8000`, what is bigger than `FLASH_SST_AREA_SIZE` [defined](https://github.com/NordicPlayground/trusted-firmware-m/pull/6/files#diff-1f387a5563b0b082bfcd63e10f69d8e6R152) as `0x4000`.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
ioannisg: @anangl i think this was copied (and adapted) from other target porting; i noticed now that it is 7 digits in Musca, and in many other portings, as well.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
ioannisg: fixing this, thanks

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
ioannisg: Yes correct, good catch.

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
ioannisg: Good point, I think this was just copied as-is from other targets.
@anangl @nvlsianpu here's the docs:
```
- ``SST_SECTORS_PER_BLOCK`` - Defines the number of contiguous SST_SECTOR_SIZE
  to form a logical block in the filesystem.
```

I think we can we safely assume that a logical block in the file-system could be 4kB, so same as the SECTOR size. So I will change this to 0x1 - any complaints?

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
```diff
@@ -0,0 +1,242 @@
...
+/* Secure + Non-secure image primary slot */
+#define FLASH_AREA_0_ID            (1)
+#define FLASH_AREA_0_OFFSET        (0x0)
+#define FLASH_AREA_0_SIZE          (FLASH_S_PARTITION_SIZE + \
+                                    FLASH_NS_PARTITION_SIZE)
+
+/* Not used*/
+#define FLASH_AREA_SCRATCH_ID      (FLASH_AREA_0_ID + 1)
+#define FLASH_AREA_SCRATCH_OFFSET  (FLASH_AREA_0_OFFSET + FLASH_AREA_0_SIZE)
+#define FLASH_AREA_SCRATCH_SIZE    (0)
```
oyvindronningstad: Why is it included if not used? Please explain in comment :).

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
anangl: No complaints from my side. Just a suggestion to do the same with `ITS_SECTORS_PER_BLOCK`.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
oyvindronningstad: +1 from me for changing to 8 or 4 digits.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
```diff
@@ -0,0 +1,159 @@
...
+#define S_HEAP_SIZE             (0x0001000)
+#define S_MSP_STACK_SIZE_INIT   (0x0000400)
+#define S_MSP_STACK_SIZE        (0x0000800)
+#define S_PSP_STACK_SIZE        (0x0000800)
+
+#define NS_HEAP_SIZE            (0x0001000)
+#define NS_MSP_STACK_SIZE       (0x0000400)
+#define NS_PSP_STACK_SIZE       (0x0000C00)
+
+#define SPU_FLASH_REGION_SIZE   (0x00008000)
```
oyvindronningstad, line 36: Is this define unique to us? If so, add a small comment.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
```diff
@@ -0,0 +1,159 @@
...
+
+/* This size of buffer is big enough to store an attestation
+ * token produced by initial attestation service
+ */
+#define PSA_INITIAL_ATTEST_TOKEN_MAX_SIZE   (0x250)
+
+/*
+ * SRAM SPU granularity is 8 KB on nRF5340. Alignment
```
oyvindronningstad: This comment looks out of place. Can you move it or add more context?

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
ioannisg: Yeap, done now, thanks @anangl

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
```diff
@@ -0,0 +1,159 @@
...
+#define NS_IMAGE_PRIMARY_PARTITION_OFFSET (FLASH_AREA_2_OFFSET \
+                                           + FLASH_S_PARTITION_SIZE)
+#endif /* !LINK_TO_SECONDARY_PARTITION */
+
+/* Boot partition structure if MCUBoot is used:
+ * 0x0_0000 Bootloader header
+ * 0x0_0400 Image area
+ * 0x0_FC00 Trailer
+ */
+/* IMAGE_CODE_SIZE is the space available for the software binary image.
```
oyvindronningstad, line 72: This and the below comment refer to IMAGE_CODE_SIZE which seems to be missing (or replaced by S and NS variants).

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
```diff
@@ -0,0 +1,159 @@
...
+#endif /* BL2 */
+
+#define IMAGE_S_CODE_SIZE \
+            (FLASH_S_PARTITION_SIZE - BL2_HEADER_SIZE - BL2_TRAILER_SIZE)
+#define IMAGE_NS_CODE_SIZE \
+            (FLASH_NS_PARTITION_SIZE - BL2_HEADER_SIZE - BL2_TRAILER_SIZE)
+
+/* The CMSE veneers shall be placed in an NSC region
+ * which will be placed in a secure SPU region defined
+ * in the beginning of the secure flash region.
```
oyvindronningstad, line 93: Does this mean an entire SPU region is used for NSC veneers? That seems like a waste of space, is it required to be individually controllable via the SPU?

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
```diff
@@ -0,0 +1,159 @@
...
+#else
+#define S_IMAGE_PRIMARY_PARTITION_OFFSET   (FLASH_AREA_2_OFFSET)
+#define S_IMAGE_SECONDARY_PARTITION_OFFSET (FLASH_AREA_0_OFFSET)
+#endif /* !LINK_TO_SECONDARY_PARTITION */
+#else
+#define S_IMAGE_PRIMARY_PARTITION_OFFSET (0x0)
+#endif /* BL2 */
+
+#ifndef LINK_TO_SECONDARY_PARTITION
+#define NS_IMAGE_PRIMARY_PARTITION_OFFSET (FLASH_AREA_0_OFFSET \
```
oyvindronningstad, line 60: Why no secondary partition for NS?

oyvindronningstad (commented):
Mostly nits and questions.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
ioannisg: will change to 8, then

ioannisg: changed.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
ioannisg: Good catch @oyvindronningstad , so I needed something for now, but I believe, this define should be taken from nrfx (FYI @anangl ) eventually. Will add a comment now, if this makes it more clear.

ioannisg: added.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
ioannisg: Actually this comment is wrong - good catch. We only care here to speak about the flash granularity, since we are defining flash partitions below. Will correct.

ioannisg: updated now

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
ioannisg: I kept it as is for consistency with the other region_defs.h headers. I 'll ask upstream for the reason behind this.

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
ioannisg: no difference - just missed to update this one.
fixed now

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
ioannisg: hm, isn't it implicit? these definitions are under #else /* ! BL2 */
so without MCUboot, there's probably no need for this.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
ioannisg: Yes, it is like this unfortunately. But that's for now - we can think of a better solution after the first porting. (Actually, that SPU region is used for both the image header and the veneers)

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
ioannisg: not sure, It does not seem to be required @oyvindronningstad 
I don't see this in other target headers

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
oyvindronningstad: Sorry, it wasn't obvious, I meant all these 4 lines :D. It says "not used" so I'm wondering if they can all be deleted.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
oyvindronningstad: I'd prefer it was referred to as something like `IMAGE_*_CODE_SIZE`, but you can leave it if you want to for the sake of consistency.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
oyvindronningstad: ok thanks for the explanation.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
oyvindronningstad: I see now they just use the secure plus an offset. let's leave it instead of creating our own system.

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
ioannisg: So, I've got a careful look into this. I think @nvlsianpu could explain this better than me, but what I understand is that MCUboot is not using the scratch area at all, so the area is not used. However, the defines _are_ used in the flash map of BL2 as you can see in bl2/src/flash_map.c and sysflash.h.

I kept here the /* Not used*/ comment, since this is present as-is in all other target platform flash layouts.

oyvindronningstad: Can you expand the comment a bit?
Also, is the bl2 flash_map.c included if `BL2` isn't defined?

ioannisg: on second thoughts, @oyvindronningstad i will remove these entries completely, and will only bring them in if we notice their are useful.
Thanks for insisting on this one.

ioannisg: Removed now :)

ioannisg:
@oyvindronningstad @anangl any more comments on this one?

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
```diff
@@ -0,0 +1,160 @@
...
+
+/* Non-secure regions */
+#define NS_IMAGE_PRIMARY_AREA_OFFSET \
+                        (NS_IMAGE_PRIMARY_PARTITION_OFFSET + BL2_HEADER_SIZE)
+#define NS_CODE_START   (NS_ROM_ALIAS(NS_IMAGE_PRIMARY_AREA_OFFSET))
+#define NS_CODE_SIZE    (IMAGE_NS_CODE_SIZE)
+#define NS_CODE_LIMIT   (NS_CODE_START + NS_CODE_SIZE - 1)
+
+#define NS_DATA_START   (NS_RAM_ALIAS(TOTAL_RAM_SIZE / 2))
+#define NS_DATA_SIZE    (TOTAL_RAM_SIZE / 2)
```
anangl:
```suggestion
#define NS_DATA_START   (NS_RAM_ALIAS(S_DATA_SIZE))
#define NS_DATA_SIZE    (TOTAL_RAM_SIZE - S_DATA_SIZE)
```
Wouldn't it be safer to define these two entries this way?

oyvindronningstad (approved):

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
ioannisg: @anangl i think your suggestion makes a lot of sense, I will apply.
(Didn't really thought so much about it; copied the solution used in the rest of the targets)

anangl: AN524 [uses](https://github.com/NordicPlayground/trusted-firmware-m/blob/master/platform/ext/target/mps3/an524/partition/region_defs.h#L114) the same what I suggested.

ioannisg: Applied now.
Actually, with a minor tweak, @anangl, I used 
`#define NS_DATA_START   (NS_RAM_ALIAS(S_DATA_LIMIT + 1))`

It is really weird that similar MPS2 targets have non-identical macro definition structures...

anangl (approved):

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
oyvindronningstad: Unsure, but I think that turns out wrong. NS_RAM_ALIAS adds the offset, but S_DATA_LIMIT already includes it.

    #define NS_DATA_START (S_DATA_LIMIT + 1)

ioannisg: Oh, that's probably correct @oyvindronningstad , maybe then I should revert this one.
It's really not ideal that there are no DATA_OFFSET macros for S and NS, as there are for CODE....

ioannisg: the original suggestion from @anangl would work, though:
`#define NS_DATA_START   (NS_RAM_ALIAS(S_DATA_SIZE))`
Updating.

PR #7, opened 2020-04-16
platform: nRF9160: adding partition information

ioannisg:
We are introducing the flash layout information and
region definitions for the nRF9160 target, to be used
in memory partitioning.


anangl (approved):

oyvindronningstad (commented):
It's pretty much identical with the nrf53 one. Should we create an nrf_common variant of these files to reduce the amount of duplication?

ioannisg:
> It's pretty much identical with the nrf53 one. Should we create an nrf_common variant of these files to reduce the amount of duplication?

Hm, we've discussed that right? :)
Let's do this after the first porting is done, to speed up the work at the moment

oyvindronningstad (approved):

PR #8, opened 2020-04-16
Integrate nrfx 2.1.0

anangl:
Import relevant files from nrfx 2.1.0 and provide a glue layer that adapts nrfx to be usable in TF-M as a host environment.

For the beginning, two nrfx drivers are imported: nrfx_nvmc and nrfx_uarte, as these are likely to be used in implementations of CMSIS Driver_Flash and Driver_USART. We'll add more drivers when it becomes necessary. From nrfx HAL, files for all peripherals available in nRF5340 application core or nRF9160 are imported.


platform/ext/target/nordic_nrf/common/nrfx_glue.c
```diff
@@ -0,0 +1,49 @@
...
+
+void  nrfx_critical_section_enter(void)
+{
+    __disable_irq();
+    ++in_critical_section;
+}
+
+void nrfx_critical_section_exit(void)
+{
+    --in_critical_section;
```
oyvindronningstad, line 46: Should there be an assert for underflow here?
```suggestion
    NRFX_ASSERT(in_critical_section > 0);
    --in_critical_section;
```

ioannisg (commented):
@anangl quick comment, regarding the cmake files: we've agreed (at least for the initial tf-m porting phase) to use the full target names in the cmake targets for nrf platforms, so probably we need to name the initial .cmake files accordingly

anangl:
> so probably we need to name the initial .cmake files accordingly

@ioannisg For sure, if we decide to move the files I added to `platform/ext` (they are now added to `platform/ext/target/nordic_nrf/nrf5340/` and `platform/ext/target/nordic_nrf/nrf9160/`, and I named them after those directories). But I thought that we could instead include them from `platform/ext/nrf5340pdk_nrf5340_cpuapp.cmake` and `platform/ext/nrf9160dk_nrf9160.cmake`, respectively, which are to be added yet. This way those files could be reused by other boards based on nRF SoCs.

ioannisg:
> > so probably we need to name the initial .cmake files accordingly
> 
> @ioannisg For sure, if we decide to move the files I added to `platform/ext` (they are now added to `platform/ext/target/nordic_nrf/nrf5340/` and `platform/ext/target/nordic_nrf/nrf9160/`, and I named them after those directories). But I thought that we could instead include them from `platform/ext/nrf5340pdk_nrf5340_cpuapp.cmake` and `platform/ext/nrf9160dk_nrf9160.cmake`, respectively, which are to be added yet. This way those files could be reused by other boards based on nRF SoCs.

Oh, sure, I notice now they are not the root .cmake files, but they are added one level deeper

platform/ext/target/nordic_nrf/common/nrfx_config.h
```diff
@@ -0,0 +1,57 @@
...
+
+/*
+ * For chips with TrustZone support, MDK provides CMSIS-Core peripheral
+ * accessing symbols in two flavors, with secure and non-secure base address
+ * mappings. Their names contain the suffix _S or _NS, respectively.
+ * Because nrfx HALs and drivers require these peripheral accessing symbols
+ * without any suffixes, the following macro is provided that will translate
+ * their names according to the kind of the target that is built.
+ */
+#if defined(NRF_TRUSTZONE_NONSECURE)
```
ioannisg, line 43: That's probably OK. I wonder only if we are supposed to build non-secure images i the tf-m project :)

ioannisg (commented):
@anangl i don't have any concerns with this PR, other than whether we need to bring in the whole nrfx drivers suite, or just the required parts of it.

Btw, probably you can skip importing the nrf51_erratas and nrf52_erratas headers, unless i am missing something here.

anangl:
> @anangl i don't have any concerns with this PR, other than whether we need to bring in the whole nrfx drivers suite, or just the required parts of it.

@ioannisg Only two drivers are imported now: nrfx_nvmc and nrfx_uarte. But as I wasn't sure which peripherals we might need to access, I included HALs for all available in nRF5340 or nRF9160. I could probably limit this list even more now, as it's rather unlikely that we'd ever need here things like I2S, PWM, or QDEC. I'll update this PR trying to remove more unnecessary stuff.

> Btw, probably you can skip importing the nrf51_erratas and nrf52_erratas headers, unless i am missing something here.

I can't. Those are [included](https://github.com/NordicPlayground/trusted-firmware-m/pull/8/files#diff-6e362dd54e03c3dcc3517b838dac429aR40) unconditionally by `nrf_erratas.h`.

ioannisg:
> I can't. Those are included unconditionally by nrf_erratas.h.

Oh, I see. But that should probably changed in the upstream project, no?

anangl:
> > I can't. Those are included unconditionally by nrf_erratas.h.
> 
> Oh, I see. But that should probably changed in the upstream project, no?

Yes, in MDK, which is imported into nrfx.

platform/ext/target/nordic_nrf/common/nrfx/mdk/nrf5340_application.h
```diff
@@ -0,0 +1,3462 @@
+/*
```
ioannisg, line 1: @anangl I've got a question about the vendor headers, e.g. this one:
In TF-M there seems to be a convention for the IRQ enum to be defined in a header called "platform_irq.h"
How is it going to work with this nrfx header?

anangl: These headers (`platform_irq.h`) are only included by other vendor files, like `tfm_peripherals_def.h`, so IMO we don't need to provide such header, we can just include `nrf.h` instead.

ioannisg: Yeah, or have a "dummy" platform_irq.h header that includes nrfx.h.
I'd say that's safer option - those headers might be there for historical reasons

platform/ext/target/nordic_nrf/common/nrfx_log.h
```diff
@@ -0,0 +1,135 @@
...
+ *        the needs of the host environment into which @em nrfx is integrated.
+ */
+
+/**
+ * @brief Macro for logging a message with the severity level ERROR.
+ *
+ * @param format printf-style format string, optionally followed by arguments
+ *               to be formatted and inserted in the resulting string.
+ */
+#define NRFX_LOG_ERROR(format, ...)
```
oyvindronningstad, line 60: Will this eventually use `LOG_MSG` from tfm_log.h?

anangl: Yes, I plan to add it later when I will be able to test it in action.

platform/ext/target/nordic_nrf/common/cmsis.h
```diff
@@ -0,0 +1,37 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __NORDIC_NRF_CMSIS_H__
+#define __NORDIC_NRF_CMSIS_H__
+
+#include <nrf.h>
```
ioannisg, line 35: @anangl in most portings, this header is part of 
`cmsis_core/Include` or `Device/Include` directories, and it includes
- `platform_regs.h`
- `platform_base_address.h`
- `platform_irq.h`

Although there does not seem to be any strict convention, I wonder how our cmsis.h could look like the other ones in the tree. your thoughts?

ioannisg (approved):
I suppose this integration is OK for now - might need to revisit things at a later stage

oyvindronningstad (approved):

tejlmand (approved):
lgtm.

Just a question:
> But I thought that we could instead include them from platform/ext/nrf5340pdk_nrf5340_cpuapp.cmake and platform/ext/nrf9160dk_nrf9160.cmake, respectively, which are to be added yet.

Do you expect to do this in a followup PR ?

ioannisg:
> lgtm.
> 
> Just a question:
> 
> > But I thought that we could instead include them from platform/ext/nrf5340pdk_nrf5340_cpuapp.cmake and platform/ext/nrf9160dk_nrf9160.cmake, respectively, which are to be added yet.
> 
> Do you expect to do this in a followup PR ?

There is an open task for this - yes.

platform/ext/target/nordic_nrf/common/nrfx_glue.c
anangl: Applied.

PR #9, opened 2020-04-17
Tfm platform peripheral defs

ioannisg:
Introducing peripheral_defs.h header for nrf5340 and nrf9160
This is as simple as it can be for the initial port.
Following the way the headers are populated for other targets.

### DIFF
--- a/platform/ext/target/nordic_nrf/nrf5340/tfm_peripherals_def.h
+++ b/platform/ext/target/nordic_nrf/nrf5340/tfm_peripherals_def.h
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2018-2019, Arm Limited. All rights reserved.
+ * Copyright (c) 2020, Cypress Semiconductor Corporation. All rights reserved.
+ * Copyright (c) 2020, Nordic Semiconductor ASA. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ */
+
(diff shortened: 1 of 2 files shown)


platform/ext/target/nordic_nrf/nrf5340/tfm_peripherals_def.h
```diff
@@ -0,0 +1,35 @@
...
+ * Copyright (c) 2020, Nordic Semiconductor ASA. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ */
+
+#ifndef __TFM_PERIPHERALS_DEF_H__
+#define __TFM_PERIPHERALS_DEF_H__
+
+#include "platform_irq.h"
```
anangl: We haven't added such file yet. And I think we should not do it, we should include here `<nrf.h>` instead to get `TIMER0_IRQn` and `TIMER1_IRQn` defined.

ioannisg: Yeap, correct. This is related to my question few days back about introducing a dummy platform_irq.h header, if you remember. 

Will remove

ioannisg: Fixed now

anangl (approved):

oyvindronningstad (approved):

PR #10, opened 2020-04-20
platform: nordic: implement tfm_platform_system.h API

ioannisg:
Implement the target-specific internal API, described
in tfm_platform_system.h, for nRF5340 and nrf9160 targets.

Note: for now the implementation concerns only System Reset (as in most of the upstream platforms, i.e. except Musca A and B1).
The implementation of GPIO and PIN services can be added at a later stage, if required.

### DIFF
--- a/platform/ext/target/nordic_nrf/nrf5340/services/src/tfm_platform_system.c
+++ b/platform/ext/target/nordic_nrf/nrf5340/services/src/tfm_platform_system.c
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2019, Arm Limited. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ */
+
+#include "platform/include/tfm_platform_system.h"
+#include "cmsis.h"
+
+void tfm_platform_hal_system_reset(void)
+{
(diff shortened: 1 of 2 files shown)


oyvindronningstad (approved):

platform/ext/target/nordic_nrf/nrf5340/services/src/tfm_platform_system.c
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) 2019, Arm Limited. All rights reserved.
```
anangl, line 2: Perhaps you should add ` * Copyright (c) 2020, Nordic Semiconductor ASA. All rights reserved.`.

anangl (approved):

platform/ext/target/nordic_nrf/nrf5340/services/src/tfm_platform_system.c
ioannisg: Hm, I don't want to do this yet - this file is almost identical to most other equivalent headers, so I'd keep ARM's copyright for now. (This will change if we add significant additional implementation here)

PR #12, opened 2020-04-23
Platform cmake

oyvindronningstad:
Flesh out nrf9160.cmake, nrf5340.cmake and nrf_common.cmake. Mostly copied and adapted from AN521/519's cmake files

Add cmake warning whenever files are missing.

I moved and renamed some files. Please protest if you disagree.

This should pass pass the cmake and compile stages and fail during linking, for both nrf platforms.


platform/ext/target/nordic_nrf/common/nrf_common.cmake
```diff
@@ -5,12 +5,144 @@
 #
 #-------------------------------------------------------------------------------
 
-set(NRF_COMMON_DIR ${CMAKE_CURRENT_LIST_DIR})
+set(NRF_COMMON_DIR "${CMAKE_CURRENT_LIST_DIR}")
 set(NRFX_DIR "${NRF_COMMON_DIR}/nrfx")
+set(PLATFORM_DIR "${NRF_COMMON_DIR}/../../..")
```
anangl: Is it needed to set it again? It was already set in the top board cmake list.

platform/ext/target/nordic_nrf/common/nrf_common.cmake
```diff
@@ -0,0 +1,150 @@
...
+                        "${NRFX_DIR}/drivers/src/nrfx_nvmc.c"
+                        "${NRF_COMMON_DIR}/nrfx_glue.c"
+                        #"${NRF_COMMON_DIR}/Native_Driver/spu.c"
+                        )
+endif()
+
+if (NOT DEFINED BUILD_TIME)
+  message(FATAL_ERROR "Configuration variable BUILD_TIME (true|false) is undefined!")
+elseif(BUILD_TIME)
+  embedded_include_directories(PATH "${NRFX_DIR}/hal" ABSOLUTE)
```
anangl: Remove this, please. nrfx HAL files should be included via `<hal/nrf_*.h>`.

platform/ext/target/nordic_nrf/common/nrf_common.cmake
oyvindronningstad: Removed

platform/ext/target/nordic_nrf/common/nrf_common.cmake
oyvindronningstad: Removed

oyvindronningstad:
@anangl @tejlmand @ioannisg @frkv Please review if you can since we have other PRs depending on this.

platform/ext/target/nordic_nrf/common/nrf_common.cmake
```diff
@@ -0,0 +1,149 @@
...
+      add_definitions(-DMCUBOOT_SIGN_RSA_LEN=3072)
+  endif()
+endif()
+
+embedded_include_directories(PATH "${NRF_COMMON_DIR}" ABSOLUTE)
+embedded_include_directories(PATH "${NRFX_DIR}" ABSOLUTE)
+embedded_include_directories(PATH "${NRFX_DIR}/drivers/include" ABSOLUTE)
+embedded_include_directories(PATH "${NRFX_DIR}/mdk" ABSOLUTE)
+embedded_include_directories(PATH "${PLATFORM_DIR}/../include" ABSOLUTE)
+embedded_include_directories(PATH "${NRF_COMMON_DIR}/retarget" ABSOLUTE)
```
anangl: I'm not sure if we'll actually need it. Perhaps it would be better to add this entry later, when we'll actually have this directory.

platform/ext/target/nordic_nrf/common/nrf_common.cmake
```diff
@@ -0,0 +1,149 @@
...
+embedded_include_directories(PATH "${NRF_COMMON_DIR}/Native_Driver" ABSOLUTE)
```
anangl: #5 uses `native_drivers`.

platform/ext/target/nordic_nrf/common/device_cfg.h
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * \file device_cfg.h
+ * \brief
+ * This is the default device configuration file with all peripherals
+ * defined and configured to be use via the secure and/or non-secure base
+ * address.
+ */
+
+#include <nrfx.h>
```
anangl: Is it needed here?

platform/ext/target/nordic_nrf/nrf5340/nrf5340.cmake
```diff
@@ -12,11 +12,42 @@ include("Common/CpuM33")
 
 set(NRF5340_DIR ${CMAKE_CURRENT_LIST_DIR})
 
-include("${NRF5340_DIR}/../common/common.cmake")
+embedded_include_directories(PATH "${NRF5340_DIR}/partition" ABSOLUTE)
```
anangl, line 15: We'll need also "`${NRF5340_DIR}`" in inclusion paths if `target_cfg.h` and `tfm_peripherals_def.h` are to stay separate for nRF5340 and nRF9160.

ioannisg (commented):
As per our offline discussion @oyvindronningstad could you remove this commit that moves the files into /common?

platform/ext/target/nordic_nrf/common/device_cfg.h
ioannisg: Not sure atm.
Also, I think this file might eventually be different in the nRF targets, so it should be added per-target, not in common

platform/ext/target/nordic_nrf/common/nrf_common.cmake
```diff
@@ -5,12 +5,145 @@
...
 set(NRFX_DIR "${NRF_COMMON_DIR}/nrfx")
 
+#Specify the location of platform specific build dependencies.
+if(COMPILER STREQUAL "ARMCLANG")
+    set (S_SCATTER_FILE_NAME   "${PLATFORM_DIR}/common/armclang/tfm_common_s.sct")
+    set (BL2_SCATTER_FILE_NAME "${NRF_COMMON_DIR}/armclang/nordic_nrf_bl2.sct")
+    set (NS_SCATTER_FILE_NAME  "${NRF_COMMON_DIR}/armclang/nordic_nrf_ns.sct")
+    if (DEFINED CMSIS_5_DIR)
+      # not all project defines CMSIS_5_DIR, only the ones that use it.
+      set (RTX_LIB_PATH "${CMSIS_5_DIR}/CMSIS/RTOS2/RTX/Library/ARM/RTX_V8MBN.lib")
```
ioannisg: This looks wrong @oyvindronningstad , should be MMN instead.
Not 100% sure, but i believe B stands for Baseline, M for Mainline
our SoCs are mainline ARMv8-M

platform/ext/target/nordic_nrf/common/nrf_common.cmake
```diff
@@ -5,12 +5,145 @@
...
+    endif()
+elseif(COMPILER STREQUAL "GNUARM")
+    set (S_SCATTER_FILE_NAME   "${PLATFORM_DIR}/common/gcc/tfm_common_s.ld")
+    set (BL2_SCATTER_FILE_NAME "${NRF_COMMON_DIR}/gcc/nordic_nrf_bl2.ld")
+    set (NS_SCATTER_FILE_NAME  "${NRF_COMMON_DIR}/gcc/nordic_nrf_ns.ld")
+    if (DEFINED CMSIS_5_DIR)
+      # not all project defines CMSIS_5_DIR, only the ones that use it.
+      set (RTX_LIB_PATH "${CMSIS_5_DIR}/CMSIS/RTOS2/RTX/Library/GCC/libRTX_V8MBN.a")
```
ioannisg: ...and here

ioannisg: Alternatively, this could be moved in the specific SoC .cmake file (if we think Nordic will eventually have a cortex-m23) :)

platform/ext/target/nordic_nrf/common/nrf_common.cmake
ioannisg: @anangl you're only talking about the "retarget" dir, right? The /../include seems required....

platform/ext/target/nordic_nrf/common/nrf_common.cmake
```diff
@@ -5,12 +5,145 @@
...
+                        "${NRFX_DIR}/drivers/src/nrfx_nvmc.c"
+                        "${NRF_COMMON_DIR}/nrfx_glue.c"
+                        #"${NRF_COMMON_DIR}/Native_Driver/spu.c"
+                        )
+endif()
+
+if (NOT DEFINED BUILD_TIME)
+  message(FATAL_ERROR "Configuration variable BUILD_TIME (true|false) is undefined!")
+elseif(BUILD_TIME)
+  #
```
ioannisg: MAybe some meaninful TODO here? @anangl what would be the timer driver source file?

ioannisg (changes requested):
Still reviewing it, but i 've got a small list of comments.

platform/ext/target/nordic_nrf/common/nrf_common.cmake
oyvindronningstad: Hehe :), if you do prefer it, I will move it 👍.

ioannisg: Not really - i find it unlikely we will need to account for Baseline v8-m architecture
But we need the fix i mention.

platform/ext/target/nordic_nrf/common/nrf_common.cmake
anangl: Yes, and yes.

platform/ext/target/nordic_nrf/common/nrf_common.cmake
oyvindronningstad: Thanks, good catch, I was not aware of the differences.

platform/ext/target/nordic_nrf/common/device_cfg.h
oyvindronningstad: Moved

platform/ext/target/nordic_nrf/common/nrf_common.cmake
anangl: The nrfx one is `nrfx/drivers/src/nrfx_timer.c`. But I'm not sure if it belongs here or rather to the `BUILD_NATIVE_DRIVERS` section, also if we need some additional wrapper for it. Maybe let's leave this section empty for now.

anangl:
@oyvindronningstad I'd rather left `nrfx_config_*.h` files in the `common/` directory. They are actually part of nrfx (its adaptation layer), and they will be updated (if needed) together with nrfx. It feels natural to keep all nrfx stuff in one place.

anangl (approved):
LGTM

ioannisg (approved):

ioannisg (approved):

PR #13, opened 2020-04-23
Add gcc startup files for nRF5340 and nRF9160

anangl:
Add startup files to be used by bootloader, secure, and non-secure images.

### DIFF
--- a/platform/ext/target/nordic_nrf/nrf5340/gcc/startup_nrf5340_ns.S
+++ b/platform/ext/target/nordic_nrf/nrf5340/gcc/startup_nrf5340_ns.S
@@ -1,0 +1,328 @@
+/*
+ * Copyright (c) 2009-2016 ARM Limited. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the License); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * This file is derivative of CMSIS V5.01 startup_ARMCM33.S
+ * Git SHA: 8a1d9d6ee18b143ae5befefa14d89fb5b3f99c75
+ */
+
+    .syntax unified
+    .arch   armv8-m.main
+
+    .section .vectors
+    .align  2
+    .globl  __Vectors
+__Vectors:
+    .long   Image$$ARM_LIB_STACK_MSP$$ZI$$Limit /* Top of Stack */
+
+    .long   Reset_Handler               /* Reset Handler */
+    .long   NMI_Handler                 /* NMI Handler */
+    .long   HardFault_Handler           /* Hard Fault Handler */
+    .long   MemManage_Handler           /* MPU Fault Handler */
+    .long   BusFault_Handler            /* Bus Fault Handler */
+    .long   UsageFault_Handler          /* Usage Fault Handler */
+    .long   0                           /* Reserved */
+    .long   0                           /* Reserved */
+    .long   0                           /* Reserved */
+    .long   0                           /* Reserved */
+    .long   SVC_Handler                 /* SVCall Handler */
+    .long   DebugMon_Handler            /* Debug Monitor Handler */
+    .long   0                           /* Reserved */
+    .long   PendSV_Handler              /* PendSV Handler */
+    .long   SysTick_Handler             /* SysTick Handler */
+
+  /* External Interrupts */
+    .long   FPU_IRQHandler
+    .long   0                           /* Reserved */
+    .long   0                           /* Reserved */
+    .long   0                           /* Reserved */
+    .long   0                           /* Reserved */
+    .long   CLOCK_POWER_IRQHandler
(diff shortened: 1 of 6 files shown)


oyvindronningstad (approved):
Thanks!

anangl:
Added "gcc" in titles.

anangl:
Corrected stack pointer linker symbols in startup files to be used by bootloader.

oyvindronningstad:
@ioannisg can you take a look? I need it for #12.

ioannisg:
> @ioannisg can you take a look? I need it for #12.

Yeap, doing a pass here after revising #5 - hang on

platform/ext/target/nordic_nrf/nrf5340/gcc/startup_nrf5340_ns.S
```diff
@@ -0,0 +1,333 @@
...
+__Vectors:
+    .long   Image$$ARM_LIB_STACK_MSP$$ZI$$Limit /* Top of Stack */
+
+    .long   Reset_Handler               /* Reset Handler */
+    .long   NMI_Handler                 /* NMI Handler */
+    .long   HardFault_Handler           /* Hard Fault Handler */
+    .long   MemManage_Handler           /* MPU Fault Handler */
+    .long   BusFault_Handler            /* Bus Fault Handler */
+    .long   UsageFault_Handler          /* Usage Fault Handler */
+    .long   SecureFault_Handler         /* Secure Fault Handler */
```
ioannisg: @anangl I think this is wrong here. you should not be needing the secure fault handler

platform/ext/target/nordic_nrf/nrf9160/gcc/startup_nrf9160_ns.S
```diff
@@ -0,0 +1,319 @@
...
+__Vectors:
+    .long   Image$$ARM_LIB_STACK_MSP$$ZI$$Limit /* Top of Stack */
+
+    .long   Reset_Handler               /* Reset Handler */
+    .long   NMI_Handler                 /* NMI Handler */
+    .long   HardFault_Handler           /* Hard Fault Handler */
+    .long   MemManage_Handler           /* MPU Fault Handler */
+    .long   BusFault_Handler            /* Bus Fault Handler */
+    .long   UsageFault_Handler          /* Usage Fault Handler */
+    .long   SecureFault_Handler         /* Secure Fault Handler */
```
ioannisg: same issue here

ioannisg (changes requested):

ioannisg:
> > @ioannisg can you take a look? I need it for #12.
> 
> Yeap, doing a pass here after revising #5 - hang on

Reviewing this atm, for now I have a question about the secure fault vector entry

platform/ext/target/nordic_nrf/nrf5340/gcc/startup_nrf5340_ns.S
anangl: Right. I'll remove those entries from _ns files.

anangl:
Removed SecureFault_Handler entries from *_ns files.

ioannisg (changes requested):
@anangl really sorry - didn't manage to post this comment: you need to remove also all IRQ vectors that refer to secure-only peripherals from the _ns.S files.

And I guess, the reverse needs to be done as well, e.g. explicitly non-secure peripheral IRQ lines cannot be present in the secure startup vector

EDIT: this is how it is done for other targets, AFAICS

anangl:
@ioannisg I removed the handler entries for secure-only peripherals from *_ns files.

> And I guess, the reverse needs to be done as well, e.g. explicitly non-secure peripheral IRQ lines cannot be present in the secure startup vector

This concerns only GPIOTE1 in both chips. But I'm not sure if I should remove those handlers. One could theoretically want to use those, at some initial stage when both VTOR_S and VTOR_NS are configured to the same value. Or do I misunderstand that?

ioannisg:
> This concerns only GPIOTE1 in both chips. But I'm not sure if I should remove those handlers. One could theoretically want to use those, at some initial stage when both VTOR_S and VTOR_NS are configured to the same value. Or do I misunderstand that?

Sure, I was thinking it's only GPIOTE1. So the story seems to be as follows (@oyvindronningstad to comment):

During the initial stage, VTOR_NS holds the reset value (I guess 0) while VTOR_S is set - by boot code -  to the address in secure partition where the vector table is located (e.g. i Zephyr this would be the start of slot0_s, which is 0x10000, i.e. after boot partition). So VTOR_NS is never pointing to the same value as VTOR_S, except for the very very first part of boot, where interrupts should anyway be disabled. That said, I think it is safe to remove GPIOTE1 from the secure startup file, but I guess it does not harm to keep. 

In absence of strong opinion, I 'd just do what other targets are doing, regarding this.

ioannisg (approved):
Looks OK now; it would be nice to clarify what we need to do with GPIOTE1_Handlers, but this is not a blocker for this PR, imo.

oyvindronningstad:
> So VTOR_NS is never pointing to the same value as VTOR_S, except for the very very first part of boot, where interrupts should anyway be disabled. That said, I think it is safe to remove GPIOTE1 from the secure startup file, but I guess it does not harm to keep.
> 
> In absence of strong opinion, I 'd just do what other targets are doing, regarding this.

+1

anangl:
> So VTOR_NS is never pointing to the same value as VTOR_S, except for the very very first part of boot, where interrupts should anyway be disabled

That's why I used the word "theoretically".

> That said, I think it is safe to remove GPIOTE1 from the secure startup file, but I guess it does not harm to keep.

There might be one case when having this vector set to the default handler instead of 0 could be useful - when VTOR_NS for whatever reason was not configured properly and the GPIOTE1 interrupt occured, this handler could help in catching the issue.
As it is harmless, I would opt for keeping those handlers.

PR #14, opened 2020-05-04
platform: nordic_nrf: Add CMSIS USART driver

anangl:
Add an implementation of the CMSIS USART driver API that uses
the nrfx_uarte driver underneath. Plus corresponding configuration
files for nRF5340 and nRF9160, and related translation layer that
activates relevant parts of nrfx.

### DIFF
--- a/platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
+++ b/platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
@@ -1,0 +1,351 @@
+/*
+ * Copyright (c) 2013-2019 Arm Limited. All rights reserved.
+ * Copyright (c) 2020 Nordic Semiconductor ASA. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the License); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <Driver_USART.h>
+#include <RTE_Device.h>
+#include <nrfx_uarte.h>
+
+#ifndef ARG_UNUSED
+#define ARG_UNUSED(arg)  (void)arg
+#endif
+
+#define ARM_USART_DRV_VERSION  ARM_DRIVER_VERSION_MAJOR_MINOR(2, 2)
+
+#if RTE_USART0 || RTE_USART1 || RTE_USART2 || RTE_USART3
+
+static const ARM_DRIVER_VERSION DriverVersion = {
+    ARM_USART_API_VERSION,
+    ARM_USART_DRV_VERSION
+};
+
+static const ARM_USART_CAPABILITIES DriverCapabilities = {
+    .asynchronous = 1,
+};
+
+typedef struct {
+    const nrfx_uarte_t   uarte;
+    const nrfx_uarte_config_t *initial_config;
+    size_t               tx_count;
+    size_t               rx_count;
+    nrf_uarte_config_t   hal_cfg;
+    nrf_uarte_baudrate_t baudrate;
+} UARTx_Resources;
+
+static ARM_DRIVER_VERSION ARM_USART_GetVersion(void)
+{
+    return DriverVersion;
+}
+
+static ARM_USART_CAPABILITIES ARM_USART_GetCapabilities(void)
+{
+    return DriverCapabilities;
+}
+
+static int32_t ARM_USARTx_Initialize(ARM_USART_SignalEvent_t cb_event,
+                                     UARTx_Resources *uart_resources)
+{
+    ARG_UNUSED(cb_event);
+
+    nrfx_err_t err_code = nrfx_uarte_init(&uart_resources->uarte,
+                                          uart_resources->initial_config,
+                                          NULL);
+    if (err_code != NRFX_SUCCESS) {
+        return ARM_DRIVER_ERROR_BUSY;
+    }
+
+    uart_resources->tx_count = 0;
+    uart_resources->rx_count = 0;
+    uart_resources->hal_cfg  = uart_resources->initial_config->hal_cfg;
+    uart_resources->baudrate = uart_resources->initial_config->baudrate;
+    return ARM_DRIVER_OK;
+}
+
+static int32_t ARM_USARTx_Uninitialize(UARTx_Resources *uart_resources)
+{
+    nrfx_uarte_uninit(&uart_resources->uarte);
+    return ARM_DRIVER_OK;
+}
+
+static int32_t ARM_USARTx_PowerControl(ARM_POWER_STATE state,
+                                       UARTx_Resources *uart_resources)
+{
+    ARG_UNUSED(uart_resources);
+
+    switch (state) {
+    case ARM_POWER_FULL:
+        /* Nothing to be done */
+        return ARM_DRIVER_OK;
+
+    case ARM_POWER_OFF:
+    case ARM_POWER_LOW:
+    default:
+        return ARM_DRIVER_ERROR_UNSUPPORTED;
+    }
+}
+
+static int32_t ARM_USARTx_Send(const void *data, uint32_t num,
+                               UARTx_Resources *uart_resources)
+{
+    nrfx_err_t err_code = nrfx_uarte_tx(&uart_resources->uarte, data, num);
+    if (err_code == NRFX_ERROR_BUSY) {
+        return ARM_DRIVER_ERROR_BUSY;
+    } else if (err_code != NRFX_SUCCESS) {
+        return ARM_DRIVER_ERROR;
+    }
+
+    uart_resources->tx_count = num;
+    return ARM_DRIVER_OK;
+}
+
+static int32_t ARM_USARTx_Receive(void *data, uint32_t num,
+                                  UARTx_Resources *uart_resources)
+{
+    nrfx_err_t err_code = nrfx_uarte_rx(&uart_resources->uarte, data, num);
+    if (err_code == NRFX_ERROR_BUSY) {
+        return ARM_DRIVER_ERROR_BUSY;
+    } else if (err_code != NRFX_SUCCESS) {
+        return ARM_DRIVER_ERROR;
+    }
+
+    uart_resources->rx_count = num;
+    return ARM_DRIVER_OK;
+}
+
+static int32_t ARM_USART_Transfer(const void *data_out, void *data_in,
+                                  uint32_t num)
+{
+    ARG_UNUSED(data_out);
+    ARG_UNUSED(data_in);
+    ARG_UNUSED(num);
+
+    return ARM_DRIVER_ERROR_UNSUPPORTED;
+}
+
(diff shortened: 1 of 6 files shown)


platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
```diff
@@ -0,0 +1,341 @@
...
+#endif
+
+#define ARM_USART_DRV_VERSION  ARM_DRIVER_VERSION_MAJOR_MINOR(2, 2)
+
+static const ARM_DRIVER_VERSION DriverVersion = {
+    ARM_USART_API_VERSION,
+    ARM_USART_DRV_VERSION
+};
+
+static const ARM_USART_CAPABILITIES DriverCapabilities = {
```
oyvindronningstad, line 37: Just curious: Which of these will we eventually support?

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
```diff
@@ -0,0 +1,341 @@
...
+     *           covered in the switch.*/
+    }
+}
+
+int32_t ARM_USARTx_Send(const void *data, uint32_t num,
+                        UARTx_Resources *uart_resources)
+{
+    nrfx_err_t err_code = nrfx_uarte_tx(&uart_resources->uarte, data, num);
+    if (err_code != NRFX_SUCCESS) {
+        return ARM_DRIVER_ERROR_BUSY;
```
oyvindronningstad, line 107: This might hide more serious errors, like NRFX_ERROR_INTERNAL.

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    uart_resources->tx_count = num;
```
oyvindronningstad, line 112: Will `uart_resources->tx_count` always be 0 before this? Otherwise, should this be `+=`?

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
```diff
@@ -0,0 +1,341 @@
...
+}
+
+int32_t ARM_USARTx_Control(uint32_t control, uint32_t arg,
+                           UARTx_Resources *uart_resources)
+{
+    if ((control & ARM_USART_CONTROL_Msk) != ARM_USART_MODE_ASYNCHRONOUS) {
+        return ARM_DRIVER_ERROR_UNSUPPORTED;
+    }
+
+    nrfx_uarte_config_t *config = &uart_resources->config;
```
oyvindronningstad: If the function exits early, this config will not match the actual configuration since the uninit/init didn't run. WIll that be a problem?

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
anangl: I used here what is currently used by all other platforms (and what is actually just copied from the driver template). TF-M code does not seem to be checking any of these capabilities. I thought about signalling RTS and CTS flow control as available, but the problem is that these two are separate capabilities here and our hardware could not support an RTS-only flow control, which is defined by the Driver_USART API.
I'll leave it as it is now.

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
anangl: This is what `ARM_USARTx_GetTxCount` returns and the description for that function [says](https://github.com/NordicPlayground/trusted-firmware-m/blob/master/platform/ext/driver/Driver_USART.h#L252) it should return the "number of data items transmitted". Other platforms return the number of bytes sent in the last transaction (i.e. call to `ARM_USART_Send`). I used the same approach.

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
anangl: Right. Will change this.

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
anangl: I thought about it. It should not be, as it is actually not described how the function should behave when only some of requested changes are successful. But using a local copy of the `config` structure seems to be safer. I'll rework it.

anangl:
Dropped [DNM] commits and rebased on master.

ioannisg:
> Dropped [DNM] commits and rebased on master.

Strill in draft mode, AFAICS, @anangl

anangl:
Refactored the code a bit, mostly `ARM_USARTx_Control()`, trying to address @oyvindronningstad's comments.
Added also a fix for an issue in `nrfx_glue.h` revealed while building this driver with the updated nrfx.

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
oyvindronningstad: You could create a macro and use it everywhere? I see initialize is the same way.

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    switch (control & ARM_USART_PARITY_Msk) {
+        case ARM_USART_PARITY_NONE:
+            hal_cfg.parity = NRF_UARTE_PARITY_EXCLUDED;
+            break;
+
+        case ARM_USART_PARITY_EVEN:
+            hal_cfg.parity = NRF_UARTE_PARITY_INCLUDED;
+            hal_cfg.paritytype = NRF_UARTE_PARITYTYPE_EVEN;
```
oyvindronningstad, line 207: I get a compile error here for nrf9160. Seems it doesn't have UARTE_CONFIG_PARITYTYPE_Msk.

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
```diff
@@ -0,0 +1,363 @@
...
+            break;
+
+        case ARM_USART_PARITY_ODD:
+            hal_cfg.parity = NRF_UARTE_PARITY_INCLUDED;
+            hal_cfg.paritytype = NRF_UARTE_PARITYTYPE_ODD;
```
oyvindronningstad, line 212: And here

oyvindronningstad (changes requested):

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
anangl: Right, I tested only on nRF5340. My bad. Will fix it promptly.

anangl: Fixed.

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
oyvindronningstad: Should ARM_USARTx_Initialize() get the same treatment?

anangl: No, the only error in `ARM_USARTx_Initialize` can be that there is an attempt to initialize an already initialized nrfx_uarte driver instance.
In `ARM_USARTx_Send` and `ARM_USARTx_Receive`, the driver can be busy with a previous transfer (`ARM_DRIVER_ERROR_BUSY` is returned then) or some other error can occur, like invalid buffer passed (not placed in RAM) or an internal nrfx_uarte driver error, then `ARM_DRIVER_ERROR` is returned.

oyvindronningstad (approved):

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
```diff
@@ -0,0 +1,369 @@
...
+    ARG_UNUSED(uart_resources);
+
+    switch (state) {
+    case ARM_POWER_OFF:
+    case ARM_POWER_LOW:
+        return ARM_DRIVER_ERROR_UNSUPPORTED;
+    case ARM_POWER_FULL:
+        /* Nothing to be done */
+        return ARM_DRIVER_OK;
+    /* default:  The default is not defined intentionally to force the
```
ioannisg: Several drivers define the default case to force the function return DRIVER_ERROR_PARAMETER - I wonder if you thought of it an chose to comply with targets that do nothing about it.

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
```diff
@@ -0,0 +1,369 @@
...
+{
+    return DriverVersion;
+}
+
+ARM_USART_CAPABILITIES ARM_USART_GetCapabilities(void)
+{
+    return DriverCapabilities;
+}
+
+int32_t ARM_USARTx_Initialize(ARM_USART_SignalEvent_t cb_event,
```
ioannisg: @anangl why these functions are not defined as static?

ioannisg (commented):
Looks fine to me.
I have a question about the driver functions not been defined with local scope, I wonder if I am missing something here.

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
```diff
@@ -0,0 +1,369 @@
...
+{
+    nrfx_err_t err_code = nrfx_uarte_tx(&uart_resources->uarte, data, num);
+    if (err_code == NRFX_ERROR_BUSY) {
+        return ARM_DRIVER_ERROR_BUSY;
+    } else if (err_code != NRFX_SUCCESS) {
+        return ARM_DRIVER_ERROR;
+    }
+
+    uart_resources->tx_count = num;
+    return NRFX_SUCCESS;
```
ioannisg: I wonder whether this is intentional or overlooked: some function return codes are taken from nrfX, such as NRFX_SUCCESS, while in other cases TF-M error code definitions are used.

anangl: Overlooked. Should be `ARM_DRIVER_OK`. Will fix it.

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
anangl: Too much relaying on the driver template. Obviously, they should be `static`. I realized that while preparing the Flash driver.

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_USART.c
anangl: I just took the approach that has been used by majority of platforms in this regard. But now I see that in Flash drivers `ARM_DRIVER_UNSUPPORTED` is mostly used in this case. This code actually seems to be a better fit than `ARM_DRIVER_ERROR_PARAMETER`. And this is also what is used in driver templates in CMSIS. I will switch to that one.

anangl:
- marked all functions that don't need to be available externally as `static`
- simplified initialization of `DriverCapabilities`
- refactored `ARM_USARTx_PowerControl`
- corrected values returned by `ARM_USARTx_Send` and `ARM_USARTx_Receive`

ioannisg (approved):
thanks for the updates @anangl

ioannisg:
@oyvindronningstad would you want/need to take another look here?

oyvindronningstad (approved):
LGTM

PR #16, opened 2020-05-07
Tfm platform cfg peripherals (nrf5340)

ioannisg:
This pull request is now ready for review, after merging #5 .
The PR does the following
- defines and implements the SPM HAL API for configuring peripheral security attribution (`platform_cfg.h` and `platform_cfg.c`)
the following API is implemented
  - initial PERIPHERAL configuration (resetting all peripherals to NS, same for GPIOS and DPPIC)
  - APIs to set a specific peripheral to Secure or Non-Secure

Unlike the other ARM vendor platforms, we cannot support APIs for setting privilege/non-privilege modes on peripheral address spaces via SPU.

The commit implements also an internal SPU "driver" API to support the target_cfg requirements.

### DIFF
--- a/platform/ext/target/nordic_nrf/common/native_drivers/spu.c
+++ b/platform/ext/target/nordic_nrf/common/native_drivers/spu.c
@@ -151,3 +151,37 @@ void spu_regions_flash_config_non_secure_callable(uint32_t start_addr,
 		FLASH_NSC_REGION_FROM_ADDR(start_addr),
 		1 /* Lock */);
 }
+
+void spu_peripheral_config_secure(uint32_t periph_base_addr, bool periph_lock)
+{
+	/* Determine peripheral ID */
+	const uint8_t periph_id = NRFX_PERIPHERAL_ID_GET(periph_base_addr);
+
+	/* ASSERT checking that this is not an explicit Non-Secure peripheral */
+	NRFX_ASSERT((NRF_SPU->PERIPHID[periph_id].PERM &
+		SPU_PERIPHID_PERM_SECUREMAPPING_Msk) !=
+		(SPU_PERIPHID_PERM_SECUREMAPPING_NonSecure <<
+			SPU_PERIPHID_PERM_SECUREMAPPING_Pos));
+
+	nrf_spu_peripheral_set(NRF_SPU, periph_id,
+		1 /* Secure */,
+		1 /* Secure DMA */,
+		periph_lock);
+}
+
+void spu_peripheral_config_non_secure(uint32_t periph_base_addr, bool periph_lock)
+{
+	/* Determine peripheral ID */
+	const uint8_t periph_id = NRFX_PERIPHERAL_ID_GET(periph_base_addr);
+
+	/* ASSERT checking that this is not an explicit Secure peripheral */
+	NRFX_ASSERT((NRF_SPU->PERIPHID[periph_id].PERM &
+		SPU_PERIPHID_PERM_SECUREMAPPING_Msk) !=
+		(SPU_PERIPHID_PERM_SECUREMAPPING_Secure <<
+			SPU_PERIPHID_PERM_SECUREMAPPING_Pos));
+
+	nrf_spu_peripheral_set(NRF_SPU, periph_id,
+		0 /* Non-Secure */,
+		0 /* Non-Secure DMA */,
+		periph_lock);
+}
--- a/platform/ext/target/nordic_nrf/common/native_drivers/spu.h
+++ b/platform/ext/target/nordic_nrf/common/native_drivers/spu.h
@@ -77,4 +77,58 @@ void spu_regions_sram_config_non_secure( uint32_t start_addr, uint32_t limit_add
  */
 void spu_regions_flash_config_non_secure_callable(uint32_t start_addr, uint32_t limit_addr);
 
+/**
+ * \brief Restrict access to peripheral to secure
+ *
+ *  Configure a device peripheral to be accessible from Secure domain only.
+ *
+ * \param periph_base_addr peripheral base address
+ *                         (must correspond to a valid peripheral ID)
+ * \param periph_lock Variable indicating whether to lock peripheral security
+ *
+ * \note
+ * - peripheral shall not be a Non-Secure only peripheral
+ * - DMA transactions are configured as Secure
+ */
+void spu_peripheral_config_secure(uint32_t periph_base_addr, bool periph_lock);
+
+/**
+ * Configure a device peripheral to be accessible from Non-Secure domain.
+ *
+ * \param periph_base_addr peripheral base address
+ *                         (must correspond to a valid peripheral ID)
+ * \param periph_lock Variable indicating whether to lock peripheral security
+ *
+ * \note
+ * - peripheral shall not be a Secure-only peripheral
+ * - DMA transactions are configured as Non-Secure
+ */
+void spu_peripheral_config_non_secure(u32_t periph_base_addr, bool periph_lock);
+
+/**
+ * Configure DPPI channels to be accessible from Non-Secure domain.
+ *
+ * \param dppi_lock Variable indicating whether to lock DPPI channel security
+ *
+ * \note all channels are configured as Non-Secure
+ */
+static inline void spu_dppi_config_non_secure(bool dppi_lock)
+{
+    nrf_spu_dppi_config_set(NRF_SPU, 0, 0x0, dppi_lock);
+}
+
+/**
+ * Configure GPIO pins to be accessible from Non-Secure domain.
+ *
+ * \param port_number GPIO Port number
+ * \param gpio_lock Variable indicating whether to lock GPIO port security
+ *
+ * \note all pins are configured as Non-Secure
+ */
+static inline void spu_gpio_config_non_secure(uint8_t port_number,
+    bool gpio_lock)
+{
+    nrf_spu_gpio_config_set(NRF_SPU, port_number, 0x0, gpio_lock);
+}
+
 #endif
--- a/platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
+++ b/platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
@@ -193,3 +193,58 @@ int32_t spu_init_cfg(void)
     return ARM_DRIVER_OK;
 }
 
+int32_t spu_periph_init_cfg(void)
+{
+	/* Peripheral configuration */
+	spu_peripheral_config_non_secure((u32_t)NRF_FPU, false);
+	spu_peripheral_config_non_secure((u32_t)NRF_OSCILLATORS, false);
+	spu_peripheral_config_non_secure((u32_t)NRF_RESET, false);
+	spu_peripheral_config_non_secure((u32_t)NRF_SPIM0, false);
+	spu_peripheral_config_non_secure((u32_t)NRF_SPIM1, false);
+	spu_peripheral_config_non_secure((u32_t)NRF_SPIM4, false);
+	spu_peripheral_config_non_secure((u32_t)NRF_SAADC, false);
+	spu_peripheral_config_non_secure((u32_t)NRF_TIMER0, false);
+	spu_peripheral_config_non_secure((u32_t)NRF_TIMER1, false);
+	spu_peripheral_config_non_secure((u32_t)NRF_TIMER2, false);
+	spu_peripheral_config_non_secure((u32_t)NRF_RTC0, false);
+	spu_peripheral_config_non_secure((u32_t)NRF_RTC1, false);
+	spu_peripheral_config_non_secure((u32_t)NRF_WDT0, false);
+	spu_peripheral_config_non_secure((u32_t)NRF_COMP, false);
+	spu_peripheral_config_non_secure((u32_t)NRF_EGU0, false);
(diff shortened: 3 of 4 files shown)


platform/ext/target/nordic_nrf/common/native_drivers/spu.c
```diff
@@ -151,3 +151,47 @@ void spu_regions_flash_config_non_secure_callable(uint32_t start_addr,
...
+	/* ASSERT checking that this is not an explicit Non-Secure peripheral */
+	NRFX_ASSERT((NRF_SPU->PERIPHID[periph_id].PERM &
+		SPU_PERIPHID_PERM_SECUREMAPPING_Msk) !=
+		(SPU_PERIPHID_PERM_SECUREMAPPING_NonSecure <<
+			SPU_PERIPHID_PERM_SECUREMAPPING_Pos));
+
+	nrf_spu_peripheral_set(NRF_SPU, periph_id,
+		1 /* Secure */,
+		1 /* Secure DMA */,
+		lock /* Lock */);
```
oyvindronningstad: `periph_lock`. You should be able to compile this now that #5 is merged.

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
```diff
@@ -77,4 +77,46 @@ void spu_regions_sram_config_non_secure( uint32_t start_addr, uint32_t limit_add
...
+ * - DMA transactions are configured as Non-Secure
+ * - region lock is passed as an argument
+ */
+void spu_peripheral_config_non_secure(u32_t periph_base_addr, bool periph_lock);
+
+/*
+ * Configure DPPI channels to be accessible from Non-Secure domain.
+ *
+ * Notes:
+ * - all channels are configred as Non-Secure
```
oyvindronningstad: typo configred -> configured

oyvindronningstad: Also below

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
```diff
@@ -77,4 +77,46 @@ void spu_regions_sram_config_non_secure( uint32_t start_addr, uint32_t limit_add
  */
 void spu_regions_flash_config_non_secure_callable(uint32_t start_addr, uint32_t limit_addr);
 
+/**
+ * \brief Restict access to peripheral to secure
```
oyvindronningstad: Typo Restict -> Restrict

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
```diff
@@ -77,4 +77,46 @@ void spu_regions_sram_config_non_secure( uint32_t start_addr, uint32_t limit_add
...
+ *
+ *  Configure a device peripheral to be accessible from Secure domain only.
+ *
+ * \note
+ * - peripheral base address must be a valid peripheral ID
```
oyvindronningstad: Some of these points could be `\param` descriptions instead.

platform/ext/target/nordic_nrf/nrf5340/target_cfg.h
```diff
@@ -45,6 +45,23 @@ struct memory_region_limits {
  */
 int32_t spu_init_cfg(void);
 
+/**
+ * \brief Configures the Security Protection Unit (peripheral security).
```
oyvindronningstad: "Security" should be "System". Otherwise this is just a suggestion I find more readable.
```suggestion
 * \brief Configures peripheral permissions via the System Protection Unit.
```

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
anangl: The parameter is currently named `periph_lock`. I guess it would be better to rename it to `lock`.
Same story in `spu_peripheral_config_non_secure`.

oyvindronningstad (changes requested):
Mostly nits

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
anangl: The peripheral ID is something different than its base address. I think this comment is a bit confusing.

platform/ext/target/nordic_nrf/nrf5340/target_cfg.h
```diff
@@ -45,6 +45,23 @@ struct memory_region_limits {
...
+ *
+ * \return  Returns error code.
+ */
+int32_t spu_periph_init_cfg(void);
```
anangl, line 58: Shouldn't the return value be rather of type `enum tfm_plat_err_t`?
Same goes for `spu_init_cfg`.

anangl (changes requested):

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
```diff
@@ -151,3 +151,47 @@ void spu_regions_flash_config_non_secure_callable(uint32_t start_addr,
...
+}
+
+void spu_dppi_config_non_secure(bool dppi_lock)
+{
+	nrf_spu_dppi_config_set(NRF_SPU, 0, 0xFFFFFFFF, dppi_lock);
+}
+
+void spu_gpio_config_non_secure(bool gpio_lock)
+{
+	nrf_spu_gpio_config_set(NRF_SPU, 0, 0xFFFFFFFF, gpio_lock);
```
anangl: And what with P1 on nRF5340?

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
ioannisg: Yeap, that's correct, this will help detecting trivial errors.
This is fixed now. Also a similar error in a function below.
I am also removing the inline comment - since the input is now parameterized.

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
ioannisg: Fixed, thanks.

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
ioannisg: thanx, fixed.

ioannisg: Btw this was copy pasted , and seems ALL other targets in the upstream tree have this typo. This says something about the upstream code quality.

platform/ext/target/nordic_nrf/nrf5340/target_cfg.h
ioannisg: Nice suggestion, really. Updated.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
ioannisg: fixed.

platform/ext/target/nordic_nrf/nrf5340/target_cfg.h
ioannisg: Updated also the spu_init_cfg documentation to match this new suggestion

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
ioannisg: yeap, did that

ioannisg: Right. So I re-wrote to: "must correspond to valid peripheral ID"

platform/ext/target/nordic_nrf/nrf5340/target_cfg.h
ioannisg: No, I prefer to leave the imt32_t because this is how the other targets implement these functions (or better: `mpc_init_cfg()` and `ppc_init_cfg()`)

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
ioannisg: Good point, this is how we have this at the moment in downstream SPM. 
@oyvindronningstad FYI: maybe we should leave P1 to secure domain in general??

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
```diff
@@ -77,4 +77,50 @@ void spu_regions_sram_config_non_secure( uint32_t start_addr, uint32_t limit_add
...
+ * - DMA transactions are configured as Non-Secure
+ */
+void spu_peripheral_config_non_secure(u32_t periph_base_addr, bool periph_lock);
+
+/**
+ * Configure DPPI channels to be accessible from Non-Secure domain.
+ *
+ * \note
+ * - all channels are configured as Non-Secure
+ * - region lock is passed as an argument
```
oyvindronningstad: Can be \param

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -193,3 +193,59 @@ int32_t spu_init_cfg(void)
     return ARM_DRIVER_OK;
 }
 
+int32_t spu_periph_init_cfg(void)
+{
+	/*
+	 * Grant Non-Secure access to nRF peripherals that are not Secure-only
```
oyvindronningstad: I suggest moving the comment to the header file.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
oyvindronningstad: It might be clearer if the ports are configurable, and everything is NS by default. Either that, or clearly explain it in the header file.

oyvindronningstad (commented):
A few more things

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
anangl: Why not take the port number as the parameter of `spu_gpio_config_non_secure()`?
BTW this function, as well as `spu_dppi_config_non_secure`, could have `__STATIC_INLINE` implementation in `spu.h`. The generated code would be probably shorter.

ioannisg: Yeap, these are all good suggestions. So:
- I will pass the port number as parameter to the public API function
- I will set P1 as well to NS  (by default) during startup configuration
- I 'll use `static inline` for these functions and "implement" the in the header

platform/ext/target/nordic_nrf/nrf5340/target_cfg.h
anangl: Well, they are called from `tfm_spm_hal_init_isolation_hw()` and their return codes are anyway translated to `TFM_PLAT_ERR_*` ones. AN539 is an exception in this regard as it uses `tfm_plat_err_t` directly.
My point is that `ARM_DRIVER_OK` returned by those functions seems a bit artificial, as it is defined for CMSIS drivers and it requires inclusion of `Driver_Common.h`. Hence my comment. But I can of course live with it. Just wanted to explain what I meant.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
ioannisg: Updated now, @oyvindronningstad @anangl

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
ioannisg: Yeap, I can do that.

ioannisg: fixed now

oyvindronningstad (approved):

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
ioannisg: fixed as well

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
```diff
@@ -77,4 +77,58 @@ void spu_regions_sram_config_non_secure( uint32_t start_addr, uint32_t limit_add
...
+void spu_peripheral_config_non_secure(u32_t periph_base_addr, bool periph_lock);
+
+/**
+ * Configure DPPI channels to be accessible from Non-Secure domain.
+ *
+ * \param dppi_lock Variable indicating whether to lock DPPI channel security
+ *
+ * \note all channels are configured as Non-Secure
+ */
+static inline void spu_dppi_config_non_secure(bool dppi_lock)
```
anangl, line 115: I guess this won't work with armclang, as it seems to use `__inline` instead of `inline`. Use `__STATIC_INLINE`.

anangl: Well, I'm probably wrong in the above, as we depend on C99 anyway, so there is no difference. Sorry for the noise.

ioannisg: Yeap, looks like we do.
And `static inline` is used extensively as is in the code base...

anangl (approved):

PR #17, opened 2020-05-11
platform: nordic_nrf: Add CMSIS Flash driver

anangl:
Add an implementation of the CMSIS Flash driver API that uses
the nrfx_nvmc driver underneath. Plus corresponding entries
in configuration files for nRF5340 and nRF9160, and a related
translation layer that activates the relevant part of nrfx.


platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_Flash.c
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    memcpy(data, (const void *)addr, cnt);
+    return ARM_DRIVER_OK;
+}
+
+static int32_t ARM_Flash_ProgramData(uint32_t addr, const void *data,
+                                     uint32_t cnt)
+{
+    if (!nrfx_is_word_aligned((const void *)addr)) {
```
oyvindronningstad: Check cnt as well?

anangl:
Rebased on master.

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_Flash.c
anangl: Added.

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_Flash.c
```diff
@@ -0,0 +1,176 @@
...
+
+    memcpy(data, (const void *)addr, cnt);
+    return ARM_DRIVER_OK;
+}
+
+static int32_t ARM_Flash_ProgramData(uint32_t addr, const void *data,
+                                     uint32_t cnt)
+{
+    /* Only aligned writes of full 32-bit words are allowed. */
+    if (!nrfx_is_word_aligned((const void *)addr) || (cnt % sizeof(uint32_t))) {
```
oyvindronningstad: It looks strange to check cnt with `%` when you are using `nrfx_is_word_aligned` in the same line.
If you don't want to use `nrfx_is_word_aligned` with a non-address, you could do 
```suggestion
    if (!nrfx_is_word_aligned((const void *)addr)
     || !nrfx_is_word_aligned((const void *)addr + cnt)) {
```

anangl: Would the following:
```
if ((addr % sizeof(uint32_t)) || (cnt % sizeof(uint32_t))) {
```
look better? I find `nrfx_is_word_aligned((const void *)addr + cnt)` somewhat artificial.

oyvindronningstad: LGTM

anangl: Corrected.

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_Flash.c
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    memcpy(data, (const void *)addr, cnt);
+    return ARM_DRIVER_OK;
+}
+
+static int32_t ARM_Flash_ProgramData(uint32_t addr, const void *data,
+                                     uint32_t cnt)
+{
+    /* Only aligned writes of full 32-bit words are allowed. */
```
ioannisg, line 118: @anangl i wonder if it would be better to copy these into dedicated function, such as 
`is_write_aligned`, is `sector_aligned`
I saw this in some flash drivers of other targets.

ioannisg (approved):
Looks ok to me.
I have a comment / question regarding refactoring a check into a private function, similar to other platforms.

oyvindronningstad (approved):

platform/ext/target/nordic_nrf/common/cmsis_drivers/Driver_Flash.c
anangl: I don't think it would make the source code more readable or the generated code better. I'd prefer to keep it as it is.

ioannisg: alright :)

PR #19, opened 2020-05-12
Remove unneeded #include

oyvindronningstad:
We don't have that file - it's a platform-internal file.

This fix is needed for certain configurations.

### DIFF
--- a/test/suites/core/non_secure/core_ns_positive_testsuite.c
+++ b/test/suites/core/non_secure/core_ns_positive_testsuite.c
@@ -20,7 +20,6 @@
 #include "tfm_veneers.h"
 #endif /* TFM_PSA_API */
 #ifdef TFM_ENABLE_IRQ_TEST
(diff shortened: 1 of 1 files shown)


interface/include/log/tfm_log_raw.h
```diff
@@ -35,7 +35,9 @@ extern "C" {
  *                      %c - character
  *                      %% - the '%' symbol
  */
-int tfm_log_printf(const char *fmt, ...);
+int printf(const char *fmt, ...);
+
+#define tfm_log_printf(...) printf(__VA_ARGS__)
```
anangl: I think that instead of doing such fix we should rather define `NRFX_ASSERT(x)` as `assert(x)`.

anangl: I prepared #23 with such change.

oyvindronningstad:
We can just bring this in as part of #35.

PR #20, opened 2020-05-12
platform: nrf: Add plat_test, including nrf_board.h files per board.

oyvindronningstad:
Support one button and one LED.
TIMER0 is used by secure, TIMER1 is used by non-secure.
Both are configured as non-secure as of now.


platform/ext/target/nordic_nrf/common/plat_test.c
```diff
@@ -0,0 +1,115 @@
...
+}
+
+void tfm_plat_test_secure_timer_start(void)
+{
+    timer_start(NRF_TIMER0);
+}
+
+void tfm_plat_test_secure_timer_stop(void)
+{
+    timer_start(NRF_TIMER0);
```
ioannisg: timer stop?

platform/ext/target/nordic_nrf/common/plat_test.c
```diff
@@ -0,0 +1,115 @@
...
+}
+
+void tfm_plat_test_non_secure_timer_start(void)
+{
+    timer_start(NRF_TIMER1);
+}
+
+void tfm_plat_test_non_secure_timer_stop(void)
+{
+    timer_start(NRF_TIMER1);
```
ioannisg: timer stop?

platform/ext/target/nordic_nrf/common/plat_test.c
oyvindronningstad: Thanks, fixed.

platform/ext/target/nordic_nrf/common/plat_test.c
```diff
@@ -0,0 +1,116 @@
...
+
+#define TIMER_RELOAD_VALUE (16*1024*1024)
+#define USERLED_MASK       (1UL)
+
+static bool initialized = false;
+
+static void gpio_init(void)
+{
+    nrf_gpio_cfg(BUTTON1_PIN, NRF_GPIO_PIN_DIR_INPUT,
+                NRF_GPIO_PIN_INPUT_DISCONNECT, BUTTON1_PULL,
```
anangl: You need to connect the input buffer to be able to read the state of the pin.
You could use `nrf_gpio_cfg_input()` here to shorten the whole thing.

platform/ext/target/nordic_nrf/common/plat_test.c
```diff
@@ -0,0 +1,116 @@
...
+                NRF_GPIO_PIN_S0S1, NRF_GPIO_PIN_NOSENSE);
+    nrf_gpio_cfg(LED1_PIN, NRF_GPIO_PIN_DIR_INPUT,
```
anangl: This pin should be configured as output. You can use `nrf_gpio_cfg_output()` for this.

platform/ext/target/nordic_nrf/boards/nrf9160dk_nrf9160/nrf_board.h
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+#ifndef NRF_BOARD_H__
+#define NRF_BOARD_H__
+
+#define BUTTON1_PIN         (2UL)
+#define BUTTON1_ACTIVE_HIGH (0UL)
+#define BUTTON1_PULL        (NRF_GPIO_PIN_PULLUP)
+#define LED1_PIN            (8UL)
+#define LED1_ACTIVE_HIGH    (0UL)
```
anangl: LEDs are active high in nRF9160 DK.

platform/ext/target/nordic_nrf/boards/nrf9160dk_nrf9160/nrf_board.h
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+#ifndef NRF_BOARD_H__
+#define NRF_BOARD_H__
+
+#define BUTTON1_PIN         (2UL)
+#define BUTTON1_ACTIVE_HIGH (0UL)
+#define BUTTON1_PULL        (NRF_GPIO_PIN_PULLUP)
+#define LED1_PIN            (8UL)
```
anangl: LED1 is on P0.02, P0.08 is connected to button 1 in nRF9160 DK.

platform/ext/target/nordic_nrf/common/plat_test.c
```diff
@@ -0,0 +1,116 @@
...
+{
+    return USERLED_MASK;
+}
+
+static void timer_start(NRF_TIMER_Type * TIMER)
+{
+    nrf_timer_mode_set(TIMER, NRF_TIMER_MODE_TIMER);
+    nrf_timer_bit_width_set(TIMER, NRF_TIMER_BIT_WIDTH_24);
+    nrf_timer_frequency_set(TIMER, NRF_TIMER_FREQ_16MHz);
+    nrf_timer_cc_set(TIMER, NRF_TIMER_CC_CHANNEL0, TIMER_RELOAD_VALUE);
```
anangl, line 85: As the timer is configured to use 24 bits, this will effectively set the compare value to 0 (the reload value is 2^24 currently). It will work, but in case the reload value is increased a bit, the effect might be be surprising.
Why not use the timer in 32-bit mode?

platform/ext/target/nordic_nrf/boards/nrf5340pdk_nrf5340_cpuapp/nrf_board.h
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (c) 2017-2020 Arm Limited. All rights reserved.
```
anangl: Was this line intentionally placed in this file?

platform/ext/target/nordic_nrf/common/plat_test.c
```diff
@@ -0,0 +1,116 @@
...
+    nrf_timer_task_trigger(TIMER, NRF_TIMER_TASK_CLEAR);
+    nrf_timer_event_clear(TIMER, NRF_TIMER_EVENT_COMPARE0);
+    nrf_timer_int_enable(TIMER, NRF_TIMER_INT_COMPARE0_MASK);
+
+    nrf_timer_task_trigger(TIMER, NRF_TIMER_TASK_START);
+}
+
+static void timer_stop(NRF_TIMER_Type * TIMER)
+{
+    nrf_timer_event_clear(TIMER, NRF_TIMER_EVENT_COMPARE0);
```
ioannisg: @anangl pls, check that the sequence of timer actions is the proper one?

platform/ext/target/nordic_nrf/common/plat_test.c
oyvindronningstad: I remember at some point not all timers supported 32 bit, but I found nothing like that for 91. Changing to 32bit.

oyvindronningstad:
@anangl Thanks a lot for the review! Should be fixed now.

platform/ext/target/nordic_nrf/common/plat_test.c
ioannisg: This does not look fixed, @oyvindronningstad , maybe GH messes it up for me

oyvindronningstad: Looks like I fixed the wrong line 🙃. Will fix properly.

platform/ext/target/nordic_nrf/common/plat_test.c
```diff
@@ -0,0 +1,113 @@
...
+    if (!initialized) gpio_init();
+
+    return (nrf_gpio_pin_out_read(LED1_PIN) == LED1_ACTIVE_HIGH);
+}
+
+void tfm_plat_test_set_led_status(uint32_t status)
+{
+    if (!initialized) gpio_init();
+
+    nrf_gpio_pin_write(LED1_PIN, status == LED1_ACTIVE_HIGH);
```
anangl: I'm afraid this won't work like that. See how the `tfm_plat_test_set_led_status()` function is used:
https://github.com/NordicPlayground/trusted-firmware-m/blob/1a02aa34328394331391284dc94dfc4e6f286b8a/test/test_services/tfm_core_test/tfm_ss_core_test.c#L209-L214
You'll need to involve `USERLED_MASK` here.

BTW Perhaps it would be better to rename the macro to `LED1_ACTIVE_LEVEL`?

platform/ext/target/nordic_nrf/common/plat_test.c
anangl: In general, the timer should be stopped first, then the event and interrupt mask should be cleared. Though, for the ONESHOT mode that is used here, there will be actually no difference.

platform/ext/target/nordic_nrf/common/plat_test.c
oyvindronningstad: I think it should work since I've defined userled_mask to be 1. Well, maybe I should mask the status for good measure. Will do that.

anangl: The thing is that the code I pointed out does not use `userled_mask`, though it probably should.

oyvindronningstad: Yep, I was thinking
```suggestion
    status &= USERLED_MASK;
    nrf_gpio_pin_write(LED1_PIN, status == LED1_ACTIVE_HIGH);
```

platform/ext/target/nordic_nrf/common/plat_test.c
oyvindronningstad: I'll switch the order.

BTW, I do int_disable(), do I need to clear the pending interrupt as well?

platform/ext/target/nordic_nrf/common/plat_test.c
oyvindronningstad: Should be fixed properly now.

anangl (approved):

platform/ext/target/nordic_nrf/common/plat_test.c
anangl: I don't know, actually. Is there a scenario in which the timer interrupt would be prevented from being serviced?

ioannisg: @anangl  I think TF-M follows standard RTOS principles with interrupt priorities, e.g. Faults and SVCs have higher priority than HW interrupts, but i guess the timer can be prioritized higher than other HW IRQs. 
@oyvindronningstad an exception here could be the crypto IRQs, but i've manage to dig deeper on this.

oyvindronningstad: Any suggestions on changes?

This file is just used for testing. Maybe it's ok as it is?

ioannisg: should be fine for now, I guess

ioannisg (approved):

PR #24, opened 2020-05-14
Platform nrf5340 spm hal

ioannisg:
SPM HAL for nRF5340

Notes
- For now, SPU Handler does not print detailed information regarding the errors
- tfm_spm_hal_confgiure_default_isolation is not implemented, because I have not figured out
  what to do with PRIV/nPRIV access policy on peripherals. Can be left for future PR.

EDIT : rebased onto master, after upmerge is done and dependencies are merged


platform/ext/target/nordic_nrf/common/native_drivers/spu.h
```diff
@@ -21,6 +21,8 @@
 #include <stdint.h>
 #include <stdbool.h>
 
+#include "Driver_Common.h"
```
anangl: Why is it needed here? This file is for CMSIS drivers, so including it here, in a native driver header, looks a bit artificial.

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
```diff
@@ -0,0 +1,278 @@
...
+ * Copyright (c) 2018-2020, Arm Limited. All rights reserved.
+ * Copyright (c) 2020, Nordic Semiconductor ASA. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ */
+
+#include <stdio.h>
+#include "cmsis.h"
+#include "platform/include/tfm_spm_hal.h"
```
anangl: `platform/include/` is in inclusion paths, so it's not needed to specify this part here.

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
```diff
@@ -0,0 +1,278 @@
...
+#include "target_cfg.h"
+#include "spu.h"
+#include "mpu_armv8m_drv.h"
+#include "region_defs.h"
+#include "secure_utilities.h"
+
+/* Get address of memory regions to configure MPU */
+extern const struct memory_region_limits memory_regions;
+
+struct mpu_armv8m_dev_t dev_mpu_s = { MPU_BASE };
```
anangl: Why is it not `static`?
And is it supposed in some future to be used by more functions than `tfm_spm_mpu_init`? If not, it could me moved there.

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
```diff
@@ -0,0 +1,278 @@
...
+}
+
+uint32_t tfm_spm_hal_get_ns_MSP(void)
+{
+    return *((uint32_t *)memory_regions.non_secure_code_start);
+}
+
+uint32_t tfm_spm_hal_get_ns_entry_point(void)
+{
+    return *((uint32_t *)(memory_regions.non_secure_code_start+ 4));
```
anangl: Wouldn't `((uint32_t *)memory_regions.non_secure_code_start)[1]` be more readable (with `...[0]` in `_get_ns_MSP`)? And actually it should be `const uint32_t *`.

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
ioannisg: Well, I thought it is the best way to take the ARM driver enums defined, e.g. ARM_DRIVER_OK,
any alternatives?

ioannisg (commented):
thanks for the review comments, @anangl , so they all make sense AFAICS; but I've just copied from the other spm_hal.c implementations available. I'll apply them, though.

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
ioannisg: i think it can be made static.
Yes, it should be used for the function that I have not yet implemented :), see TODO, FIXME

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
ioannisg: alright

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
anangl: I guess it's about `ARM_DRIVER_OK` not defined in `target_cfg.c`. But the question is if you in fact need it there? It seems that `spu_init_cfg()` and `spu_periph_init_cfg()` could use other error codes, not the CMSIS driver ones.

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
ioannisg: again, that was copy-pasted. Seems the same to me, I think either way the developer needs to know that the reset handler is the second word of the table...

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
ioannisg: fixed now.

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
ioannisg: I made this static now.

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
ioannisg: I made this pointer _const_, by kept the +4, not only for uniformity with the other targets, but also since +4 makes it more obvious that we are pointing to the cortex-m ResetHandler.

ioannisg: Hope this is fine with you (at the end is perhaps a matter of taste)

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
ioannisg: I would like to preserve uniformity, and consistency in `spm_hal.c` function `tfm_spm_hal_init_isolation_hw()` (which checks that the _init_cfg() functions return ARM_DRIVER_OK), across the various supported platforms.

If spu_init_cfg() and spu_periph_init_cfg() would return other error codes (e.g. Nordic specific), then the spm_hal would need to read these other error codes.

If this is preferable, let me know. I don't care so much about the error codes here as they are not propagated in higher levels.

As a side note, since the MPC (and PPC) _are actually_ CMSIS drivers, I would think that in the "near" future, our spu.c chould also be made a CMSIS driver and then it would make sense to return the ARM_DRIVER_OK

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
```diff
@@ -0,0 +1,278 @@
...
+{
+    return *((uint32_t *)memory_regions.non_secure_code_start);
+}
+
+uint32_t tfm_spm_hal_get_ns_entry_point(void)
+{
+    return *((const uint32_t *)(memory_regions.non_secure_code_start + 4));
+}
+
+enum tfm_plat_err_t tfm_spm_hal_set_secure_irq_priority(int32_t irq_line,
```
oyvindronningstad: Note: After the upmerge, replace all instances of `int32_t irq_line` with `IRQn_Type irq_line`

oyvindronningstad (approved):

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
ioannisg: Yeap, saw it as well. 
thanks for reminding

ioannisg: done now, @oyvindronningstad

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
anangl: > I would like to preserve uniformity, and consistency in `spm_hal.c` function `tfm_spm_hal_init_isolation_hw()` (which checks that the _init_cfg() functions return ARM_DRIVER_OK), across the various supported platforms.
> 
> If spu_init_cfg() and spu_periph_init_cfg() would return other error codes (e.g. Nordic specific), then the spm_hal would need to read these other error codes.

AN539 uses `enum tfm_plat_err_t` for those. To me it looks more adequate.
https://github.com/NordicPlayground/trusted-firmware-m/blob/94f4fb3510acaabf430781d45aa40267fa9ac34d/platform/ext/target/mps2/an539/spm_hal.c#L36-L53

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
anangl: Sure, it's fine. Since you decided to add `const` to the pointer, you should do the same in `tfm_spm_hal_get_ns_MSP()`, for consistency.

anangl (approved):
I'm still not entirely convinced regarding inclusion of `Driver_Common.h` and the values returned by `spu_init_cfg()` and `spu_periph_init_cfg()`, but it's actually not a big issue and it should not block merging of this PR.

platform/ext/target/nordic_nrf/common/native_drivers/spu.h
ioannisg: @anangl  AN539 is the only one that returns TFM_PLAT_ERR codes.
But then I think you might be right here. Right now nRF SPU is not a CMSIS "driver" (yet) so it might make more sense to use the TFM Platform Error codes directly.

As i mentioned above, my only intention was to preserve uniformity, but since there is already one platform deviating, it does not harm to have nRF ones doing the same.

Will update based on your proposal

ioannisg:
> I'm still not entirely convinced regarding inclusion of `Driver_Common.h` and the values returned by `spu_init_cfg()` and `spu_periph_init_cfg()`, but it's actually not a big issue and it should not block merging of this PR.

True, however, as i mentioned above, I'll accept your feedback before merging this PR.

PR #27, opened 2020-05-15
platform: nordic_nrf: flash_layout: remove macros for non-BL2 builds

ioannisg:
It turns out that we do not need the FLASH_AREA_0_
macro definitions for TF-M builds without BL2. These
macros are not used, they are not present in most of
the other platform targets.

### DIFF
--- a/platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
+++ b/platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
@@ -67,7 +67,7 @@
 /* Flash layout info for BL2 bootloader */
 #define FLASH_BASE_ADDRESS                  (0x00000000)
 
-#ifdef BL2
+
 /* Offset and size definitions of the flash partitions that are handled by the
  * bootloader. The image swapping is done between IMAGE_PRIMARY and
  * IMAGE_SECONDARY, SCRATCH is used as a temporary storage during image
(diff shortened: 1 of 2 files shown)


ioannisg:
It does not seem these are needed for non-BL2 builds.
I copied them from the nxp PR (where they are still present after the merge). But they are not present in any other target flash layout. I think this PR fixes the error of the strate imgtool script fails, as @anangl mentioned offline.

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
```diff
@@ -67,7 +67,7 @@
 /* Flash layout info for BL2 bootloader */
 #define FLASH_BASE_ADDRESS                  (0x00000000)
 
-#ifdef BL2
+
```
oyvindronningstad, line 70: Nit: This line isn't added for the 91 file.

oyvindronningstad (approved):

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
ioannisg: fixed thanks

oyvindronningstad:
To rebase onto new master:

    git fetch [nordic]
    git rebase -i before-upmerge-1 --onto [nordic]/master

ioannisg:
> To rebase onto new master:
> 
> ```
> git fetch [nordic]
> git rebase -i before-upmerge-1 --onto [nordic]/master
> ```

Thanks. Possible to do, also, with just `rebase --interactive` and picking the original commits in the PR

anangl (approved):

ioannisg (commented):
Rebased after the upmerge

PR #28, opened 2020-05-18
Fix missing bin file when building tests.

oyvindronningstad:
Always create bin file into same dir as axf file.


ioannisg (approved):
Looks ok to me.
This fix makes the build on Regression cmake targets, as far as I can see

ioannisg:
@tejlmand can you take a look?

oyvindronningstad:
Split into two commits, one that fixes upstream, and one that fixes our patch where we added hex and elf generation.

ioannisg:
@oyvindronningstad  do you anticipate that the upstream fix will eventually come through a future upmerge? If so, can you mark the commit title accordingly, to help out resolving possible conficts?

Otherwise, that's OK with me - just adapt both commit titles to follow a natural format with ":" 's
:)

oyvindronningstad:
> @oyvindronningstad do you anticipate that the upstream fix will eventually come through a future upmerge? If so, can you mark the commit title accordingly, to help out resolving possible conficts?

I don't know, not in the near future, but we could start opening PRs early when we have tested more. I think I'll leave it for now.
 
> Otherwise, that's OK with me - just adapt both commit titles to follow a natural format with ":" 's
> :)

Done, they are quite cmake-wide commits, so I just added "cmake:".

ioannisg:
@oyvindronningstad you can rebase this one now

secure_fw/CMakeLists.txt
```diff
@@ -331,8 +331,15 @@ function(set_up_secure_fw_build)
...
+	#Collect executables to common location: <build_dir>/install/outputs/
+	if (DEFINED SECURE_AXF_DIR_PREFIX)
+		set(MY_BINARY_DIR ${SECURE_AXF_DIR_PREFIX})
+	else()
+		set(MY_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
+	endif()
+
 	#Generate binary file from executable
-	compiler_generate_binary_output(${EXE_NAME})
+	compiler_generate_binary_output(${EXE_NAME} ${MY_BINARY_DIR})
```
tejlmand, line 342: If I understand the TF-M build system correctly, then it is wrong to directly target `${MY_BINARY_DIR}` for output.

Everything should be placed in the normal build path, and then be copied from their to the install folder, as part of the install target. That is what would happen on `make/ninja install`

cmake/Common/CompilerArmClangCommon.cmake
```diff
@@ -184,13 +184,13 @@ function(compiler_merge_library)
 	target_sources(${_MY_PARAMS_DEST} PRIVATE ${_MY_PARAMS_LIBS})
 endfunction()
 
-function(compiler_generate_binary_output TARGET)
+function(compiler_generate_binary_output TARGET DIR)
```
tejlmand, line 187: if doing this the correct way, as part of install target, see https://github.com/NordicPlayground/trusted-firmware-m/pull/28#discussion_r428023553, then this `DIR` patch shouldn't be needed

secure_fw/CMakeLists.txt
```diff
@@ -342,10 +342,10 @@ function(set_up_secure_fw_build)
 	compiler_generate_binary_output(${EXE_NAME} ${MY_BINARY_DIR})
 
 	#Generate intel hex file from axf
-	compiler_generate_hex_output(${EXE_NAME})
+	compiler_generate_hex_output(${EXE_NAME} ${MY_BINARY_DIR})
```
tejlmand, line 345: same as other comments.
Should happen on executing install target.

secure_fw/CMakeLists.txt
```diff
@@ -342,10 +342,10 @@ function(set_up_secure_fw_build)
...
 
 	#Generate elf file from axf
-	compiler_generate_elf_output(${EXE_NAME})
+	compiler_generate_elf_output(${EXE_NAME} ${MY_BINARY_DIR})
```
tejlmand, line 348: same as other comments.
Should happen on executing install target.

tejlmand (changes requested):

secure_fw/CMakeLists.txt
oyvindronningstad: I'm pretty sure I'm following convention. AFAIU `MY_BINARY_DIR` and `SECURE_AXF_DIR_PREFIX` are used to divert the build output, not the install output. When I build it with ConfigRegression.cmake which sets `CORE_TEST` , the axf is not in its regular build location (secure_fw), but is instead in unit_test.

From secure_fw/CMakeLists.txt:
```
if(CORE_TEST)
	set(SECURE_AXF_DIR_PREFIX "${CMAKE_BINARY_DIR}/unit_test/")
	set_target_properties(${EXE_NAME} PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${SECURE_AXF_DIR_PREFIX})
	embedded_set_target_link_defines(TARGET ${EXE_NAME} DEFINES "TFM_PARTITION_TEST_CORE")
endif()

...

#Collect executables to common location: <build_dir>/install/outputs/
if (DEFINED SECURE_AXF_DIR_PREFIX)
	set(MY_BINARY_DIR ${SECURE_AXF_DIR_PREFIX})
else()
	set(MY_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
endif()

...

install(FILES ${MY_BINARY_DIR}/${EXE_NAME}.axf
	      ${MY_BINARY_DIR}/${EXE_NAME}.bin
	      ${MY_BINARY_DIR}/${EXE_NAME}.hex # Added by us
	      ${MY_BINARY_DIR}/${EXE_NAME}.elf # Added by us
	      DESTINATION outputs/${TARGET_PLATFORM}/)
```

tejlmand: ahh, right, my mistake.
The install destination is `output/${TARGET_PLATFORM}`, not `${MY_BINARY_DIR}`.

oyvindronningstad:
@tejlmand any more comments?

ioannisg:
@tejlmand ping

tejlmand (approved):

PR #29, opened 2020-05-19
platform: nordic_nrf: Add VTOR initialization in secure startup code

anangl:
Add initialization of SCB->VTOR in the startup code for secure images
on nRF5340 and nRF9160. Other platforms do this at the beginning of
SystemInit(), but since nRF targets use the implementation of this
function from MDK, it cannot be done exactly the same way.

### DIFF
--- a/platform/ext/target/nordic_nrf/nrf5340/gcc/startup_nrf5340_s.S
+++ b/platform/ext/target/nordic_nrf/nrf5340/gcc/startup_nrf5340_s.S
@@ -244,6 +244,12 @@ Reset_Handler:
 #endif /* __STARTUP_CLEAR_BSS_MULTIPLE || __STARTUP_CLEAR_BSS */
 
     cpsid   i              /* Disable IRQs */
+
+/* Setup Vector Table Offset Register. */
+    ldr     r0, =__Vectors
+    ldr     r1, =0xE000ED08 /* SCB->VTOR */
+    str     r0, [r1]
+
     bl      SystemInit
 
     mrs     r0, control    /* Get control value */
--- a/platform/ext/target/nordic_nrf/nrf9160/gcc/startup_nrf9160_s.S
+++ b/platform/ext/target/nordic_nrf/nrf9160/gcc/startup_nrf9160_s.S
@@ -240,6 +240,12 @@ Reset_Handler:
 #endif /* __STARTUP_CLEAR_BSS_MULTIPLE || __STARTUP_CLEAR_BSS */
 
     cpsid   i              /* Disable IRQs */
+
+/* Setup Vector Table Offset Register. */
+    ldr     r0, =__Vectors
+    ldr     r1, =0xE000ED08 /* SCB->VTOR */
+    str     r0, [r1]
+
     bl      SystemInit
 
     mrs     r0, control    /* Get control value */


platform/ext/target/nordic_nrf/nrf5340/gcc/startup_nrf5340_s.S
```diff
@@ -244,6 +244,13 @@ Reset_Handler:
 #endif /* __STARTUP_CLEAR_BSS_MULTIPLE || __STARTUP_CLEAR_BSS */
 
     cpsid   i              /* Disable IRQs */
+
+/* Setup Vector Table Offset Register. */
+    ldr     r0, =__Vectors
+    ldr     r1, =#0xE000ED08 /* SCB->VTOR */
+    str     r0, [r1]
+    dsb     0xF
```
ioannisg: Changing VTOR requires ISB as well. This is because an error may occur during fetching an instruction, and we want to make sure the new vector table will apply.

ioannisg: btw, i don't thing the 0xF in the dsb instruction is needed

anangl: Perhaps I'm missing something or I just don't understand it well. ISB will flush the pipeline, so all the instructions following ISB will be fetched again. Consequently, for a few instructions the fetching will be done twice, with the old and with the new vector table. If an error was to occur during both these fetchings, the corresponding fault handler from the old vector table would be executed anyway first. And if the first fetching was successful, what's the point in trying once more? I don't see a clear benefit of having ISB here and I think it might be even confusing.

ioannisg: https://static.docs.arm.com/dai0321/a/DAI0321A_programming_guide_memory_barriers_for_m_profile.pdf

In general, your are right @anangl, I had a more careful look. Actually, if you see the attached application note, even DSB is not required, according to the implementation requirements :)  That's because Cortex-M always completes memory writes before exception entries.

However, all this depend on the assumption that the _old_ vector table was a "valid" one. Assume, for example that the "old" vector table was existing but "dummy", e.g. was just containing dummy fault handlers which were silently ignoring faults. We are now changing to the new vector table, and, in theory, we might want to re-fetch all instructions, to see if there were a fetching fault. :)

anangl: Removed those DSB. They're indeed redundant for Cortex-M processors.

platform/ext/target/nordic_nrf/nrf5340/gcc/startup_nrf5340_s.S
```diff
@@ -244,6 +244,12 @@ Reset_Handler:
 #endif /* __STARTUP_CLEAR_BSS_MULTIPLE || __STARTUP_CLEAR_BSS */
 
     cpsid   i              /* Disable IRQs */
+
+/* Setup Vector Table Offset Register. */
+    ldr     r0, =__Vectors
+    ldr     r1, =#0xE000ED08 /* SCB->VTOR */
```
ioannisg: I think the "#" needs to be omitted (pseudo-ldr does not need it,  i think). 
MOV instructions need #.

ioannisg: You can define the VTOR as a constant. See an example from Zehyr:
```

    ldr v4, =_SCS_ICSR
    ldr v3, =_SCS_ICSR_UNPENDSV

```

...where the macros are defined (for ASM)

oyvindronningstad:
How about making a C function and calling into that?

ioannisg (approved):

platform/ext/target/nordic_nrf/nrf5340/gcc/startup_nrf5340_s.S
anangl: I know. But I didn't want to introduce new definitions just for this one line.
I just used `SCB->VTOR` as the comment to make it clear what is happening here and to make it possible to easily locate the place where VTOR is initialized for nRF targets (just by grepping for `SCB->VTOR`).

anangl:
> How about making a C function and calling into that?

@oyvindronningstad  But then we would need to add one more C file into compilation (or place the function in a file not strictly related to the startup code), just to get the address of `SCB->VTOR`. I though it was not worth it.

oyvindronningstad:
> > How about making a C function and calling into that?
> 
> @oyvindronningstad But then we would need to add one more C file into compilation, just to get the address of `SCB->VTOR`. I though it was not worth it.

Yep, just wanted to suggest it since I personally find assembly very fiddly :).

oyvindronningstad (approved):

ioannisg:
> > How about making a C function and calling into that?
> 
> @oyvindronningstad But then we would need to add one more C file into compilation (or place the function in a file not strictly related to the startup code), just to get the address of `SCB->VTOR`. I though it was not worth it.

Thinking that other vendors do this in a similar way in SystemInit, I am not sure it is worth adding a function for this. [I know we have done that in Zephyr, though, but in Zephyr we have many factors that affect VTOR, e.g. XIP, vector table relocation, etc. But this does not hold in this case, now :) ]

PR #30, opened 2020-05-20
Leave UARTE1 to Secure State when building with Secure UART1

ioannisg:


ioannisg:
@oyvindronningstad can you check how I've modified the cmake files? I think it's in accordance with other targets that enable SECURE UART1, but pls take a look.

@anangl ideally this will solve your issue with trying to log over UART for TF-M, provided that we are able to use UARTE1 for console output

oyvindronningstad (approved):
LGTM!

anangl (approved):

PR #31, opened 2020-05-20
platform: nrf5340: spm_hal: configure peripheral default isolation

ioannisg:
We implement the function to configure peripheral isolation
for peripherals belonging to (Secure) partitions. We implement
the MPU configuration, allowing the peripheral to be accessed
by non-privileged code, and configure the SPU, setting the
peripheral to be accessible by Secure domain only.

Notes
- the function is supposed to implement isolation for peripherals belonging to secure partitions.

- the function uses the MPU configuration on peripheral space, similar to what the other targets are doing. This is done so as to be able to use a peripheral from nPRIV code
- in addition, the function configures a peripheral in this list to Secure. Note that SPU cannot assign PRIV/nPRIV attribution configuration, so inevitably this function only adds Secure configuration.

### DIFF
--- a/platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
+++ b/platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
@@ -17,11 +17,25 @@
 #include "mpu_armv8m_drv.h"
 #include "region_defs.h"
 #include "secure_utilities.h"
+#include "region.h"
 
 /* Get address of memory regions to configure MPU */
 extern const struct memory_region_limits memory_regions;
 
+#if defined(CONFIG_TFM_ENABLE_MEMORY_PROTECT)
 static struct mpu_armv8m_dev_t dev_mpu_s = { MPU_BASE };
+
+#define MPU_REGION_VENEERS              0
+#define MPU_REGION_TFM_UNPRIV_CODE      1
+#define MPU_REGION_TFM_UNPRIV_DATA      2
+#define MPU_REGION_NS_STACK             3
+#define PARTITION_REGION_RO             4
+#define PARTITION_REGION_RW_STACK       5
+#define PARTITION_REGION_PERIPH_START   6
+#define PARTITION_REGION_PERIPH_MAX_NUM 2
+
+static uint32_t periph_num_count = 0;
+#endif /* CONFIG_TFM_ENABLE_MEMORY_PROTECT */
 
 enum tfm_plat_err_t tfm_spm_hal_init_isolation_hw(void)
 {
@@ -46,23 +60,51 @@ enum tfm_plat_err_t tfm_spm_hal_configure_default_isolation(
                   uint32_t partition_idx,
                   const struct tfm_spm_partition_platform_data_t *platform_data)
 {
-    /* FIXME we need to see how to handle this because nRF IDAU peripheral
-     * is not able to assign PRIV/nPRIV access policy configuration
+    if (!platform_data) {
+        return TFM_PLAT_ERR_INVALID_INPUT;
+    }
+
+#if defined(CONFIG_TFM_ENABLE_MEMORY_PROTECT) && (TFM_LVL != 1)
+
+    if (!tfm_is_partition_privileged(partition_idx)) {
+        struct mpu_armv8m_region_cfg_t region_cfg;
+
+        region_cfg.region_nr = PARTITION_REGION_PERIPH_START +
+                                periph_num_count;
+        periph_num_count++;
+        if (periph_num_count >= PARTITION_REGION_PERIPH_MAX_NUM) {
+            return TFM_PLAT_ERR_MAX_VALUE;
+        }
+        region_cfg.region_base = platform_data->periph_start;
+        region_cfg.region_limit = platform_data->periph_limit;
+        region_cfg.region_attridx = MPU_ARMV8M_MAIR_ATTR_DEVICE_IDX;
+        region_cfg.attr_access = MPU_ARMV8M_AP_RW_PRIV_UNPRIV;
+        region_cfg.attr_sh = MPU_ARMV8M_SH_NONE;
+        region_cfg.attr_exec = MPU_ARMV8M_XN_EXEC_NEVER;
+
+        mpu_armv8m_disable(&dev_mpu_s);
+
+        if (mpu_armv8m_region_enable(&dev_mpu_s, &region_cfg)
+            != MPU_ARMV8M_OK) {
+            return TFM_PLAT_ERR_SYSTEM_ERR;
+        }
+        mpu_armv8m_enable(&dev_mpu_s, PRIVILEGED_DEFAULT_ENABLE,
+                          HARDFAULT_NMI_ENABLE);
+    }
+#endif /* defined(CONFIG_TFM_ENABLE_MEMORY_PROTECT) && (TFM_LVL != 1) */
+
+    /* Peripheral for a (secure) partition is configured as Secure.
+     * Note that the SPU does not allow to configure PRIV/nPRIV permissions
+     * for address ranges, including peripheral address space.
      */
+    if (platform_data->periph_start != 0) {
+        spu_periph_configure_to_secure(platform_data->periph_start);
+    }
+
     return TFM_PLAT_ERR_SUCCESS;
 }
 
 #ifdef CONFIG_TFM_ENABLE_MEMORY_PROTECT
-
-#define MPU_REGION_VENEERS           0
-#define MPU_REGION_TFM_UNPRIV_CODE   1
-#define MPU_REGION_TFM_UNPRIV_DATA   2
-#define MPU_REGION_NS_STACK          3
-#define PARTITION_REGION_RO          4
-#define PARTITION_REGION_RW_STACK    5
-#define PARTITION_REGION_PERIPH      6
-
-
 REGION_DECLARE(Image$$, TFM_UNPRIV_CODE, $$RO$$Base);
 REGION_DECLARE(Image$$, TFM_UNPRIV_CODE, $$RO$$Limit);
 REGION_DECLARE(Image$$, TFM_UNPRIV_DATA, $$RW$$Base);
--- a/platform/ext/target/nordic_nrf/nrf5340/target_cfg.h
+++ b/platform/ext/target/nordic_nrf/nrf5340/target_cfg.h
@@ -39,6 +39,15 @@ struct memory_region_limits {
 };
 
 /**
+ * \brief Holds the data necessary to do isolation for a specific peripheral.
+ */
+struct tfm_spm_partition_platform_data_t
+{
+    uint32_t periph_start;
+    uint32_t periph_limit;
+};
+
+/**
  * \brief Configures memory permissions via the System Protection Unit.
  *
  * \return  Returns error code.


platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
```diff
@@ -46,9 +48,49 @@ enum tfm_plat_err_t tfm_spm_hal_configure_default_isolation(
...
+    struct mpu_armv8m_region_cfg_t region_cfg;
+#endif
+
+    if (!platform_data) {
+        return TFM_PLAT_ERR_INVALID_INPUT;
+    }
+
+#if defined(CONFIG_TFM_ENABLE_MEMORY_PROTECT) && (TFM_LVL != 1)
+    if (!privileged) {
+        region_cfg.region_nr = PARTITION_REGION_PERIPH_START +
```
oyvindronningstad, line 72: `PARTITION_REGION_PERIPH_START` and `periph_num_count` look undefined.

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
```diff
@@ -46,9 +48,49 @@ enum tfm_plat_err_t tfm_spm_hal_configure_default_isolation(
...
+                          HARDFAULT_NMI_ENABLE);
+    }
+#endif /* defined(CONFIG_TFM_ENABLE_MEMORY_PROTECT) && (TFM_LVL != 1) */
+
+    /* Peripheral for a (secure) partition is configured as Secure.
+     * Note that the SPU does not allow to configure PRIV/nPRIV permissions
+     * for address ranges, including peripheral address space.
      */
+    if (platform_data->periph_base_addr != 0) {
+        spu_periph_configure_to_secure(platform_data->periph_base_addr);
```
oyvindronningstad: Curious: I see that fex. AN521 configures MPU *and* ppc_en_secure_unpriv. What does ppc_en_secure_unpriv do that the MPU doesn't?

oyvindronningstad (changes requested):

platform/ext/target/nordic_nrf/nrf5340/target_cfg.h
```diff
@@ -38,6 +38,14 @@ struct memory_region_limits {
 #endif /* BL2 */
 };
 
+/**
+ * \brief Holds the data necessary to do isolation for a specific peripheral.
+ */
+struct tfm_spm_partition_platform_data_t
```
oyvindronningstad, line 44: Since you introduced code that uses it, I think you should add `periph_start` and `periph_limit` also.

ioannisg: why?
these are needed to configure isolation using ARM's PPC, but Nordic does not need them. Only the base address is needed by the SPU.

oyvindronningstad: They are used in lines 68 and 69 in the code you added.

ioannisg: Oh oh, right, missed that - i need them yes, thanks

ioannisg: [the SPU _does_ need only the base address, but the MPU needs the limits.]

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
```diff
@@ -21,7 +21,9 @@
 /* Get address of memory regions to configure MPU */
 extern const struct memory_region_limits memory_regions;
 
+#if defined(CONFIG_TFM_ENABLE_MEMORY_PROTECT) && (TFM_LVL != 1)
 static struct mpu_armv8m_dev_t dev_mpu_s = { MPU_BASE };
```
anangl: `dev_mpu_s` is used also by `tfm_spm_mpu_init()` which is present when only `CONFIG_TFM_ENABLE_MEMORY_PROTECT` is defined.

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
ioannisg: @oyvindronningstad this is a good question: I suspect that the PPC controller in ARM MPS2+ FPGAs has the power to assign privilege/un-privilege access permissions. Nordic SPU only assigns RWX permissions (ignoring the privilege level). 

In Cortex-m all access policies of this kind are implementing the AND combination rule.

So, to be able to access a peripheral from un-privileged level, both the MPU and PPC need to allow this.

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
ioannisg: That's fixed now, thanks.

platform/ext/target/nordic_nrf/nrf5340/target_cfg.h
ioannisg: fixed, thanks

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
oyvindronningstad: Does it work when they are defined after the usage?

ioannisg: Nope. I had to re-shuffle that. Also I had to include region.h (this came from last upmerge)
Done now.

ioannisg (commented):
@oyvindronningstad you could take a look now.
It now passes compilation. I am testing build with IPC + TFM Level2 isolation which defines 
CONFIG_TFM_ENABLE_MEMORY_PROTECT and sets TFM_LVL to 2

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
```diff
@@ -46,23 +62,53 @@ enum tfm_plat_err_t tfm_spm_hal_configure_default_isolation(
                   uint32_t partition_idx,
                   const struct tfm_spm_partition_platform_data_t *platform_data)
 {
-    /* FIXME we need to see how to handle this because nRF IDAU peripheral
-     * is not able to assign PRIV/nPRIV access policy configuration
+#if defined(CONFIG_TFM_ENABLE_MEMORY_PROTECT) && (TFM_LVL != 1)
+    bool privileged = tfm_is_partition_privileged(partition_idx);
+    struct mpu_armv8m_region_cfg_t region_cfg;
```
anangl: Why not move this into the `if (!privileged)` block where it is only used? Actually, you could also not use the `privileged` variable and get rid of this `#if` here completely.

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
```diff
@@ -46,23 +62,53 @@ enum tfm_plat_err_t tfm_spm_hal_configure_default_isolation(
...
+        region_cfg.region_attridx = MPU_ARMV8M_MAIR_ATTR_DEVICE_IDX;
+        region_cfg.attr_access = MPU_ARMV8M_AP_RW_PRIV_UNPRIV;
+        region_cfg.attr_sh = MPU_ARMV8M_SH_NONE;
+        region_cfg.attr_exec = MPU_ARMV8M_XN_EXEC_NEVER;
+
+        mpu_armv8m_disable(&dev_mpu_s);
+
+        if (mpu_armv8m_region_enable(&dev_mpu_s, &region_cfg)
+            != MPU_ARMV8M_OK) {
+            return TFM_PLAT_ERR_SYSTEM_ERR;
```
anangl, line 89: MPU will stay disabled in case of this error. Is that okay?

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
```diff
@@ -17,11 +17,27 @@
...
+#define MPU_REGION_VENEERS              0
+#define MPU_REGION_TFM_UNPRIV_CODE      1
+#define MPU_REGION_TFM_UNPRIV_DATA      2
+#define MPU_REGION_NS_STACK             3
+#define PARTITION_REGION_RO             4
+#define PARTITION_REGION_RW_STACK       5
+#define PARTITION_REGION_PERIPH_START   6
+#define PARTITION_REGION_PERIPH_MAX_NUM 2
+
+uint32_t periph_num_count = 0;
```
anangl: Why is it not static?

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
ioannisg: I assume so. This results in a system error, probably calling tfm_panic(), which processes the error properly. So that's not a problem in my opinion. That code should be locking or resetting the device anyway.

(After all, if MPU could not configure a region properly, then something is really wrong so MPU being ON would not make too much difference)

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
ioannisg: Yeap, right it can be. That was copy pasted

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
ioannisg: that's right, will fix.

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
ioannisg: I understood that this is because there are some "weird" coding style guidelines in tf-m that want us to define the variables in the top of the function

oyvindronningstad (approved):

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
```diff
@@ -17,11 +17,27 @@
...
+#include "region.h"
 
 /* Get address of memory regions to configure MPU */
 extern const struct memory_region_limits memory_regions;
 
+#if defined(CONFIG_TFM_ENABLE_MEMORY_PROTECT)
 static struct mpu_armv8m_dev_t dev_mpu_s = { MPU_BASE };
+#endif
+
+#ifdef CONFIG_TFM_ENABLE_MEMORY_PROTECT
```
anangl: You can merge these `#if` blocks into one.

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
anangl: I thought it was to be the top of a block, not necessarily a function. But if it is indeed about a function, you could always extract the block to be a separate function. Of course, it's not a big deal, I just find code that is less polluted with `#if`s a bit easier to read.

ioannisg: No big deal either, I can go with your proposal (just wanted to communicate why I had done that this way)

ioannisg: done now

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
ioannisg: done thanks

platform/ext/target/nordic_nrf/nrf5340/spm_hal.c
```diff
@@ -46,23 +60,51 @@ enum tfm_plat_err_t tfm_spm_hal_configure_default_isolation(
...
-     * is not able to assign PRIV/nPRIV access policy configuration
+    if (!platform_data) {
+        return TFM_PLAT_ERR_INVALID_INPUT;
+    }
+
+#if defined(CONFIG_TFM_ENABLE_MEMORY_PROTECT) && (TFM_LVL != 1)
+    bool privileged = tfm_is_partition_privileged(partition_idx);
+    struct mpu_armv8m_region_cfg_t region_cfg;
+
+    if (!privileged) {
```
anangl: In my earlier comment I meant to do it this way:
```suggestion
    if (!tfm_is_partition_privileged(partition_idx)) {
        struct mpu_armv8m_region_cfg_t region_cfg;
```

anangl (approved):

PR #33, opened 2020-05-22
platform: nordic_nrf: Fixes for non-secure domain builds

anangl:
A couple of corrections and additions needed for a proper building
of the non-secure image.


platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -132,7 +132,7 @@ enum tfm_plat_err_t nvic_interrupt_target_state_cfg(void)
```
ioannisg, line 132: @anangl could you also fix line 131 with SPU_IRQn?

ioannisg (commented):
Looks ok to me. Please, update also the entry with SPU_IRQn, to use the NRFX macro function

ioannisg (approved):

anangl:
Moved two more files to the secure image only source files list (I forgot about this change initially as I had it in a different commit).

oyvindronningstad (approved):

PR #34, opened 2020-05-22
platform: nordic_nrf: nrfx: Use NRF_TRUSTZONE_NONSECURE for NS domain

anangl:
When building for non-secure domain, define the corresponding macro
needed by MDK and nrfx drivers.

### DIFF
--- a/platform/ext/target/nordic_nrf/common/nrfx_config.h
+++ b/platform/ext/target/nordic_nrf/common/nrfx_config.h
@@ -54,6 +54,10 @@
 #define NRFX_UARTE3_ENABLED 1
 #endif
 
+#if (__DOMAIN_NS == 1)
+#define NRF_TRUSTZONE_NONSECURE
+#endif
+
 /*
  * For chips with TrustZone support, MDK provides CMSIS-Core peripheral
  * accessing symbols in two flavors, with secure and non-secure base address
--- a/platform/ext/target/nordic_nrf/common/nrfx/drivers/src/nrfx_nvmc.c
+++ b/platform/ext/target/nordic_nrf/common/nrfx/drivers/src/nrfx_nvmc.c
@@ -152,6 +152,22 @@ __STATIC_INLINE bool is_halfword_aligned(uint32_t addr)
     return ((addr & 0x1u) == 0u);
 }
 
+__STATIC_INLINE bool is_valid_address(uint32_t addr)
+{
+    if ((addr - NVMC_FLASH_BASE_ADDRESS) < flash_total_size_get())
+    {
+        return true;
+    }
+#if !defined(NRF_TRUSTZONE_NONSECURE)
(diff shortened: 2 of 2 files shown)


ioannisg (commented):
This is taken from the nRFx PR @anangl , right?
(The one that I approved earlier)

anangl:
> This is taken from the nRFx PR @anangl , right?
> (The one that I approved earlier)

Yes, the [DNM] commit. I planned to replace is with an import of the patch from the nrfx repo, after it is merged there.

platform/ext/target/nordic_nrf/common/nrfx_config.h
```diff
@@ -54,6 +54,10 @@
 #define NRFX_UARTE3_ENABLED 1
 #endif
 
+#if (__DOMAIN_NS == 1)
+#define NRF_TRUSTZONE_NONSECURE
```
ioannisg, line 58: Hm, I wonder, if this could be set as a cmake variable, as we do in Zephyr.
@oyvindronningstad ^^

oyvindronningstad: NRF_TRUSTZONE_NONSECURE? Sure, but I'm not sure why we want to, can you elaborate?

oyvindronningstad (approved):

platform/ext/target/nordic_nrf/common/nrfx_config.h
anangl: We want to, for example, because this macro is used also in `system_*.c` files, which are compiled without processing this file (they don't include `<nrfx.h>`, only `<nrf.h>`). Currently nothing from those files is used in non-secure images, but...

oyvindronningstad: Could they just use __DOMAIN_NS instead?

anangl: No, because they are imported from nrfx/MDK.

oyvindronningstad: Ok, I would think just adding `#include <nrfx.h>` would be cleaner, but I'm probably missing something. I will stop now :).

ioannisg: @anangl ^^ 
Also wondering if including nrfx.h could be better, pls, clarify :)

anangl: The thing is that the files that would need to include `<nrfx.h>`:
- `system_nrf5340_application.c`
- `system_nrf9160.c`

come from MDK that is imported with nrfx. It would not be good to modify them here in TF-M.

ioannisg: got it now, thanks

ioannisg (approved):

PR #35, opened 2020-05-26
Rearrange nsc

oyvindronningstad:


platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
```diff
@@ -51,8 +51,8 @@
  */
 
 /* Size of a Secure and of a Non-secure image */
-#define FLASH_S_PARTITION_SIZE                (0x30000)       /* S partition: 192 kB*/
-#define FLASH_NS_PARTITION_SIZE               (0x40000)       /* NS partition: 256 kB*/
+#define FLASH_S_PARTITION_SIZE                (0x41000)       /* S partition: 260 kB*/
+#define FLASH_NS_PARTITION_SIZE               (0x30000)       /* NS partition: 192 kB*/
```
ioannisg, line 55: I am not sure I like this @oyvindronningstad 
I want the flash layout to be compatible with zephyr default upstream partitioning, so NSPE needs to start on 0x40000

platform/ext/common/gcc/tfm_common_s.ld
```diff
@@ -56,6 +65,25 @@ SECTIONS
...
+     * this region as Non-Secure Callable.
+     */
+    .gnu.sgstubs : ALIGN(32)
+    {
+        SG_ALIGN;
+        Load$$LR$$LR_VENEER$$Base = ABSOLUTE(.);
+        *(.gnu.sgstubs*)
+    } > VENEERS AT> VENEERS
+    SG_END_ALIGN;
+    Load$$LR$$LR_VENEER$$Limit = .;
```
anangl: This value is [written](https://github.com/NordicPlayground/trusted-firmware-m/blob/cf25ed7a1f1ac14a1f7244877c4485311cb76c91/platform/ext/target/nordic_nrf/nrf5340/target_cfg.c#L51) directly to `memory_regions.veneer_limit`, so it should be probably `. -1` here (or this initialization of `memory_regions` needs to be corrected). Otherwise, configuration of the NSC region ends up in HardFault.

ioannisg:
Rebased your PR @oyvindronningstad based on latest master

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
oyvindronningstad: Well, like it or not we need the extra space for secure_fw in configurations like `ConfigRegression.cmake`

    Linking C executable unit_test/tfm_s.axf
    Memory region         Used Size  Region Size  %age Used
               FLASH:      252896 B       258 KB     95.72%
                 RAM:       81352 B       256 KB     31.03%

platform/ext/common/gcc/tfm_common_s.ld
oyvindronningstad: Thanks, fixed

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
ioannisg: OK, so we need a clean solution here
- first, any changes here need to be reflected in the comment section above, which gives the overview of the flash layout. Lines 21-44.
- then, I am wondering whether 129kB for NSPE are sufficient, e.g. for NCS applications (I know that most of upstream  Zephyr samples for nRF5340 will fit in 192kB). If 129kB is too small, I wonder if we could use some #ifdef's here and have "conditional" flash layout-

platform/ext/target/nordic_nrf/nrf9160/partition/flash_layout.h
```diff
@@ -51,7 +51,7 @@
  */
 
 /* Size of a Secure and of a Non-secure image */
-#define FLASH_S_PARTITION_SIZE                (0x30000)       /* S partition: 192 kB*/
+#define FLASH_S_PARTITION_SIZE                (0x31000)       /* S partition: 196 kB*/
```
ioannisg: Why is this different from nRF5340?

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
```diff
@@ -74,7 +74,7 @@
  * swapping.
  */
 #define FLASH_AREA_BL2_OFFSET      (0x0)
-#define FLASH_AREA_BL2_SIZE        (0x10000) /* 64 KB */
+#define FLASH_AREA_BL2_SIZE        (0xF000) /* 60 KB */
```
ioannisg: This looks sufficient - I wonder if we should update upstream Zephyr to use 60kB (0xF00) or even 48kB, which is the case for nRF52

platform/ext/target/nordic_nrf/nrf9160/partition/flash_layout.h
oyvindronningstad: Slipup, I forgot to change this one. Thanks

platform/ext/common/gcc/tfm_common_s.ld
```diff
@@ -56,6 +65,25 @@ SECTIONS
...
+    {
+        SG_ALIGN;
+        Load$$LR$$LR_VENEER$$Base = ABSOLUTE(.);
+        *(.gnu.sgstubs*)
+    } > VENEERS AT> VENEERS
+    SG_END_ALIGN;
+    Load$$LR$$LR_VENEER$$Limit = . - 1;
+
+    ASSERT ((Load$$LR$$LR_VENEER$$Limit - Load$$LR$$LR_VENEER$$Base)
+            <= CMSE_VENEER_REGION_SIZE, "Veneer region overflowed")
```
anangl: After the definition of `Load$$LR$$LR_VENEER$$Limit` has been changed, this assertion should be updated to use `<` instead of `<=`.

oyvindronningstad: Good catch!

oyvindronningstad: Fixed.

oyvindronningstad:
I pushed a change that works with the veneers XiP in the middle of the image. It wastes between 0 and 32 k, but can serve as our working example until we have something better (like RAM). Will look into the RAM change now.

platform/ext/common/gcc/tfm_common_s.ld
```diff
@@ -589,6 +598,32 @@ SECTIONS
...
+     * Place the CMSE Veneers (containing the SG instruction) after the code, in a
+     * separate 32 bytes aligned region so that the SAU can programmed to just set
+     * this region as Non-Secure Callable.
+     */
+    .gnu.sgstubs : ALIGN(32)
+    {
+        SG_ALIGN;
+        Load$$LR$$LR_VENEER$$Base = ABSOLUTE(.);
+        *(.gnu.sgstubs*)
+    } > VENEERS AT> VENEERS
```
anangl: For `CMSE_VENEER_REGION_IN_CODE`, there is no region `VENEERS` defined, so the above looks a bit weird (it should be `> FLASH` in that case).

platform/ext/common/gcc/tfm_common_s.ld
```diff
@@ -589,6 +598,32 @@ SECTIONS
...
+
+    /* This is needed because the linker seems to place the veneers *after* the
+       .gnu.sgstubs output section, so any end alignment applied inside it are
+       placed before the veneers instead of after. */
+    .sg_end_dummy :
+    {
+        SG_END_ALIGN;
+        Load$$LR$$LR_VENEER$$Limit = ABSOLUTE(. - 1);
+    } > VENEERS AT> VENEERS
```
anangl: ~~But you don't need to define another output section for it. It should be sufficient to just move `SG_END_ALIGN` out of the `.gnu.sgstubs` section.~~
EDIT: I see I was wrong in the above. Sorry for the noise.

test/suites/core/non_secure/core_ns_positive_testsuite.c
```diff
@@ -20,7 +20,6 @@
 #include "tfm_veneers.h"
 #endif /* TFM_PSA_API */
 #ifdef TFM_ENABLE_IRQ_TEST
-#include "platform_irq.h"
```
ioannisg, line 23: Well, this is not clean solution @oyvindronningstad 
I am afraid this can't be accepted upstream.
Maybe bring back an empty platform_irq.h for nRF targets (I remember I had added that during porting and was asked to remove it as not used)

platform/ext/common/gcc/tfm_common_s.ld
oyvindronningstad: I wanted to avoid `ifdef`ing this. Should I instead change it to VENEERS_REGION and define that to either FLASH or VENEERS?

anangl: Well, perhaps `#ifdef`ing here would be cleaner, however. You wouldn't need to define `SG_ALIGN` and `SG_END_ALIGN` then. The comment before the `.gnu.sgstubs` section is also not perfectly accurate for the `CMSE_VENEER_REGION_IN_CODE` code.

test/suites/core/non_secure/core_ns_positive_testsuite.c
oyvindronningstad: It's not really clean to include platform_irq.h here because it contains platform-specific stuff. And anyways, it's not used in this file.

platform/ext/common/gcc/tfm_common_s.ld
oyvindronningstad: I can ifdef just that line if you really prefer that, but I really don't want to duplicate the linker code since it's very finicky, and I want to make sure the implementations don't go out of sync with each other.

I'll also fix the comment above.

test/suites/core/non_secure/core_ns_positive_testsuite.c
ioannisg: In these cases, I would normally want this to be sent upstream as a separate PR, first. We're not doing this now, but I assume we have a plan to do this :)

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
ioannisg: It looks clean now.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
```diff
@@ -104,17 +98,25 @@
...
 
-/* CMSE Veneers region */
-#define CMSE_VENEER_REGION_START  (S_IMAGE_PRIMARY_AREA_OFFSET - CMSE_VENEER_REGION_SIZE)
+/* The CMSE veneers shall be placed in an NSC region
+ * which will be placed in a secure SPU region defined
+ * in the beginning of the secure flash region.
+ */
+#define CMSE_VENEER_REGION_SIZE     (0x400)
+#define CMSE_VENEER_REGION_START_ALIGN \
+                . = ALIGN(SPU_FLASH_REGION_SIZE) - (CMSE_VENEER_REGION_SIZE)
```
ioannisg: Are these defines only to be used by the linker?
I suppose this file is a standard header to be included in C code, so I'd rather see the #defines not containing the dots :)

ioannisg: Instead I'd like to see an absolute value of the address of the start of the veneers region, this is how it is done in other targets, right?

ioannisg (changes requested):

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
oyvindronningstad: > I'd rather see the #defines not containing the dots :)

I can change it to

    #define CMSE_VENEER_REGION_START_ALIGN (ALIGN(SPU_FLASH_REGION_SIZE) - (CMSE_VENEER_REGION_SIZE))

> Instead I'd like to see an absolute value of the address of the start of the veneers region, this is how it is done in other targets, right?

Yes, but we don't want that when we are placing the veneers in code, since we want the linker to place them at the earliest free location that satisfies the requirements. Otherwise we risk wasting even more space.

test/suites/core/non_secure/core_ns_positive_testsuite.c
oyvindronningstad: Of course. Can we assume that this is fine, and if they don't accept it upstream when we PR it, we add an empty platform_irq.h at that point?

ioannisg: sounds good

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
ioannisg: Ah, right. Then, please, document that, because this is different from other targets. Some inline comments explaining the usage of the macros would suffice.

platform/ext/common/gcc/tfm_common_s.ld
oyvindronningstad: Take a look now.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
oyvindronningstad: Take a look now.

platform/ext/common/gcc/tfm_common_s.ld
```diff
@@ -589,6 +601,44 @@ SECTIONS
...
+     * aligned region so that the SAU can programmed to just set this region as
+     * Non-Secure Callable.
+     */
+    .gnu.sgstubs : ALIGN(32)
+    {
+        . = CMSE_VENEER_REGION_START_ALIGN;
+        Load$$LR$$LR_VENEER$$Base = ABSOLUTE(.);
+        *(.gnu.sgstubs*)
+#ifdef CMSE_VENEER_REGION_IN_CODE
+    } > FLASH AT> FLASH
```
anangl: I think `AT>` is unneeded in this case.

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
```diff
@@ -34,7 +34,7 @@
 
 /* Size of nRF SPU (Nordic IDAU) regions */
 #define SPU_FLASH_REGION_SIZE   (0x00008000)
-#define SPU_SRAM_REGION_SIZE   (0x00002000)
+#define SPU_SRAM_REGION_SIZE    (0x00002000)
```
anangl, line 37: Could you add such definition for nRF9160 on the occasion (it's missing)?

anangl (approved):

platform/ext/common/gcc/tfm_common_s.ld
oyvindronningstad: Thanks, done

platform/ext/target/nordic_nrf/nrf5340/partition/region_defs.h
oyvindronningstad: done

platform/ext/common/gcc/tfm_common_s.ld
```diff
@@ -31,11 +31,23 @@ MEMORY
 #if defined(S_CODE_SRAM_ALIAS_BASE)
   CODE_RAM (rwx) : ORIGIN = S_CODE_SRAM_ALIAS_BASE, LENGTH = TOTAL_CODE_SRAM_SIZE
 #endif
-#ifndef TFM_MULTI_CORE_TOPOLOGY
+#if !defined(TFM_MULTI_CORE_TOPOLOGY) && !defined(CMSE_VENEER_REGION_IN_CODE)
   VENEERS  (rx)  : ORIGIN = CMSE_VENEER_REGION_START, LENGTH = CMSE_VENEER_REGION_SIZE
 #endif
 }
 
+/* Define this macro in the platform's region_defs.h to override the default
```
ioannisg: Can you put this into #ifdef CMSE_VENEER_REGION_IN_CODE?
I propose to decouple completely the solution when VENEERS are in code and when they are not. This is much easier to read, and it is not too much of duplication..

platform/ext/common/gcc/tfm_common_s.ld
```diff
@@ -31,11 +31,23 @@ MEMORY
 #if defined(S_CODE_SRAM_ALIAS_BASE)
   CODE_RAM (rwx) : ORIGIN = S_CODE_SRAM_ALIAS_BASE, LENGTH = TOTAL_CODE_SRAM_SIZE
 #endif
-#ifndef TFM_MULTI_CORE_TOPOLOGY
+#if !defined(TFM_MULTI_CORE_TOPOLOGY) && !defined(CMSE_VENEER_REGION_IN_CODE)
```
ioannisg: I would leave the first #ifdef as is, and then add a nested #ifdef with an explicit comment that elaborates why we do not need a liner area for VENEERS

ioannisg (changes requested):
Overall looks ok, but I've got a "request" to separate our additions from the default TFM linker organization (with respect to the veneers).

platform/ext/common/gcc/tfm_common_s.ld
oyvindronningstad: Can you please elaborate? I don't know exactly what you are asking for.

I suppose without _IN_CODE there's less of a reason for alignment since platforms fully control the placement, so adding an ifdef is fine with me.

I really don't want to duplicate the sgstubs stuff though if that's what you meant since it's so finicky, and if they go out of sync it will be a lot of work to understand any changes and apply them after the fact.

platform/ext/common/gcc/tfm_common_s.ld
oyvindronningstad: Done

platform/ext/common/gcc/tfm_common_s.ld
oyvindronningstad: I changed this somewhat, making it less verbose, take a look.

platform/ext/common/gcc/tfm_common_s.ld
```diff
@@ -32,10 +32,21 @@ MEMORY
...
+ * the rest of the code. If so, also define CMSE_VENEER_REGION_START_ALIGN and
+ * CMSE_VENEER_REGION_END_ALIGN.
+ */
+#ifndef CMSE_VENEER_REGION_IN_CODE
   VENEERS  (rx)  : ORIGIN = CMSE_VENEER_REGION_START, LENGTH = CMSE_VENEER_REGION_SIZE
 #endif
+#endif
 }
 
+#ifndef CMSE_VENEER_REGION_IN_CODE
```
ioannisg, line 46: @oyvindronningstad strictly speaking, this is only needed insind an #ifndef TFM_MULTI_CORE_TOPOLOGY

oyvindronningstad: Added an #ifdef TFM_MULTI_CORE_TOPOLOGY around it.

platform/ext/common/gcc/tfm_common_s.ld
ioannisg: Good enough for now, I am not 100% convinced, but I don't feel i should delay this more, for now

platform/ext/common/gcc/tfm_common_s.ld.template
```diff
@@ -253,6 +266,44 @@ SECTIONS
...
+     * This is needed because the linker seems to place the veneers *after* the
+     * .gnu.sgstubs output section, so any end alignment applied inside it are
+     * placed before the veneers instead of after. Also, the end alignment seems
+     * to be ignored if not done in an output section, so this cannot be placed
+     * by itself after the .gnu.sgstubs section.
+     */
+    .sg_end_dummy : ALIGN(32)
+    {
+        . = CMSE_VENEER_REGION_END_ALIGN;
+        Load$$LR$$LR_VENEER$$Limit = ABSOLUTE(. - 1);
```
ioannisg: @oyvindronningstad could you please, consider one more issue here: 
the _Limit address for the veneers' section is calculated, subtracting 1 from the current address pointer, which is different from the original implementation, where Limit is simply _start_ + sizeof (veneers section size)

That, also, appears to be the "standard" way of defining " veneer limit" addresses in TF-M; they do not consider the -1 in the calculations. See, for example, in target_cfg.c

```
    .veneer_base = (uint32_t)&REGION_NAME(Load$$LR$$, LR_VENEER, $$Base),
    .veneer_limit = (uint32_t)&REGION_NAME(Load$$LR$$, LR_VENEER, $$Limit),
```

I think this is because the veneers' region is only used in MPU configuration, not in SAU, IDAU.

oyvindronningstad: I see, wasn't aware. Updated now, thanks!

ioannisg (approved):

platform/ext/common/gcc/tfm_common_s.ld.template
anangl: We need to fix then (in a follow-up PR, I guess) this call:
https://github.com/NordicPlayground/trusted-firmware-m/blob/333c93e337a8f72fc4cf78a22e1247f4a65f6d9b/platform/ext/target/nordic_nrf/nrf5340/target_cfg.c#L185-L186
It's quite confusing that  "veneer limit" is understood differently than "non secure partition limit".

anangl (approved):

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -48,7 +59,7 @@ const struct memory_region_limits memory_regions = {
         (uint32_t)&REGION_NAME(Load$$LR$$, LR_VENEER, $$Base),
 
     .veneer_limit =
-        (uint32_t)&REGION_NAME(Load$$LR$$, LR_VENEER, $$Limit),
+        (uint32_t)&REGION_NAME(Load$$LR$$, LR_VENEER, $$Limit) - 1,
```
ioannisg: I think this is not correct, @oyvindronningstad because the veneer_limit is used in the MPU configuration, which _expects_ that the veneer_limit points to the address _right after_ the end of the section. If you see the MPU region configuration, the function subtracts 1 byte from the address, so if you subtract one more here, it will not be accurate.

(TBH, this is a bad design from the tf-m code)

For now, I guess you can do the -1 directly in the spu CFG function

oyvindronningstad: veneer_limit is used only one place, in `spu_regions_flash_config_non_secure_callable()`, but I can change it in the parameter list instead if you prefer that.

oyvindronningstad: Done

ioannisg: That's exactly what I meant to say

PR #36, opened 2020-05-28
platform: nordic_nrf: Import patch for nrfx_nvmc.c from nrfx repository (2)

anangl:
This imports commit 233c96307e0946fc44035424c410bd44ddf2d75e from
the nrfx repository at https://github.com/NordicSemiconductor/nrfx,
which introduces the following change:

nrfx_nvmc: Fix setting of program memory access mode for secure code

For secure code, the access mode needs to be set for both secure and
non-secure regions. Otherwise, attempts from secure code to write or
erase a region configured as non-secure end up with BusFault.


oyvindronningstad (approved):

ioannisg (approved):
Approving this, since this is brought from NRFX as is.
We can follow up the issues with the comments there , if needed.

PR #38, opened 2020-06-02
Reduce diff

oyvindronningstad:
From looking at the upstream PR for Ninja (https://review.trustedfirmware.org/c/trusted-firmware-m/+/4368), it seems all my changes were not needed. There are three commits that are not necessary, so this removes them to reduce the diff against upstream.


ioannisg:
What should the workflow be here? Should we first wait until the upstream PR gets merged, then we revert and do the upmerge together?
@oyvindronningstad ^^

oyvindronningstad:
> What should the workflow be here?

It's quite independent IMO. We can review and merge this. Then, when we rebase again we can remove the reverts together with the original commits. If the upstream pr gets merged, we can remove a different set of commits from our history if we like. Also, I think it's better to do the upmerge like in downstream Zephyr, which means upmerge and rebase are independent from each other as well.

ioannisg:
@tejlmand , pls, take a quick look on this revert attempt :)

ioannisg:
@tejlmand ping

ioannisg (approved):

tejlmand (approved):

oyvindronningstad:
merge?

PR #40, opened 2020-06-05
platform: nrf9160: Provide SPM HAL, enable SECURE_UART1 option

anangl:
Provide required SPM HAL functions for the nRF9160 platform,
together with the target_cfg implementation (mostly by copying
from nRF5340).
Make UARTE1 a Secure only peripheral by taking advantage of
the SECURE_UART1 option.

Further two commits contain small corrections for nRF5340.


platform/ext/target/nordic_nrf/nrf9160/target_cfg.c
```diff
@@ -0,0 +1,243 @@
...
+    reg_value |= (uint32_t)(SCB_AIRCR_WRITE_MASK | SCB_AIRCR_SYSRESETREQS_Msk);
+
+    SCB->AIRCR = reg_value;
+
+    return TFM_PLAT_ERR_SUCCESS;
+}
+
+enum tfm_plat_err_t init_debug(void)
+{
+    return TFM_PLAT_ERR_SUCCESS;
```
anangl, line 113: Should there be a compilation error issued here if an option different than `DAUTH_CHIP_DEFAULT` is used? Something like what is used in LPC55S69: https://github.com/NordicPlayground/trusted-firmware-m/blob/778b9b141a6cde2fd31c7e7825efabdcde7503f2/platform/ext/target/nxp/lpcxpresso55s69/target_cfg.c#L128-L130.

platform/ext/target/nordic_nrf/nrf9160/spm_hal.c
```diff
@@ -0,0 +1,323 @@
+/*
```
anangl: What about making this file a common one for nRF5340 and nRF9160? Currently, for both platforms it is identical.

oyvindronningstad: I think it's good idea.

ioannisg: I agree, as long as we are confident we are not going to have #ifdef PLATFORM directives there.

anangl: Moved the file to `nordic_nrf/common/`.

oyvindronningstad (approved):
LGTM

platform/ext/target/nordic_nrf/nrf5340/target_cfg.c
```diff
@@ -209,8 +209,8 @@ enum tfm_plat_err_t spu_periph_init_cfg(void)
 {
 	/* Peripheral configuration */
 	spu_peripheral_config_non_secure((uint32_t)NRF_FPU, false);
-	spu_peripheral_config_non_secure((uint32_t)NRF_OSCILLATORS, false);
-	spu_peripheral_config_non_secure((uint32_t)NRF_RESET, false);
+	spu_peripheral_config_non_secure((uint32_t)NRF_REGULATORS, false);
+	spu_peripheral_config_non_secure((uint32_t)NRF_CLOCK, false);
```
ioannisg, line 212: OK @anangl so as far as I understand from the commit message, this is not wrong but just done for uniformity with nrf9160

anangl: Yes, exactly, only for uniformity.

platform/ext/target/nordic_nrf/nrf9160/target_cfg.c
ioannisg: yes, I think so @anangl

anangl: Added an error message.

anangl:
Removed unnecessary entries that configured non-secure access to the FPU, on both nRF platforms.

ioannisg (approved):

PR #41, opened 2020-06-17
Fixes needed for RegressionIPC configuration

anangl:
Zephyr [tfm_integration samples](https://github.com/zephyrproject-rtos/zephyr/tree/master/samples/tfm_integration) use RegressionIPC configuration for building TF-M.

With the changes contained in this PR it is possible to successfully build and run those samples on both nRF5340 and nRF9160.
In order to use this branch in Zephyr, in place of its [fork of TF-M](https://github.com/zephyrproject-rtos/trusted-firmware-m/tree/master/trusted-firmware-m), some more adaptations from its trusted-firmware-m module need to be applied, see https://github.com/anangl/trusted-firmware-m/commit/c761f2efffba3bee6eb4d4a4aa786188d18f5926.

And finally, the following changes in Zephyr's board definitions are needed for building and running the tfm_integration samples with TF-M from this PR:
- [nrf5340pdk_nrf5340](https://github.com/anangl/zephyr/commit/c82e113d1e797b9c01e52bd157ca2085721536be)
- [nrf9160dk_nrf9160](https://github.com/anangl/zephyr/commit/19dba803664201b0bb2c76aef7a88b93cbe4eb63)

### DIFF
--- a/platform/ext/common/gcc/tfm_common_s.ld
+++ b/platform/ext/common/gcc/tfm_common_s.ld
@@ -602,6 +602,29 @@ SECTIONS
 
 #endif /* TFM_LVL != 1 */
 
+    .ER_TFM_CODE : ALIGN(4)
+    {
+        *(.text*)
+
+        KEEP(*(.init))
+        KEEP(*(.fini))
+
+
+        /* .ctors */
+        *crtbegin.o(.ctors)
+        *crtbegin?.o(.ctors)
+        *(EXCLUDE_FILE(*crtend?.o *crtend.o) .ctors)
+        *(SORT(.ctors.*))
+        *(.ctors)
+
+        /* .dtors */
+         *crtbegin.o(.dtors)
+         *crtbegin?.o(.dtors)
+         *(EXCLUDE_FILE(*crtend?.o *crtend.o) .dtors)
+         *(SORT(.dtors.*))
+         *(.dtors)
+    } > FLASH
+
 #ifndef TFM_MULTI_CORE_TOPOLOGY
     /*
      * Place the CMSE Veneers (containing the SG instructions) in a correctly
@@ -640,28 +663,8 @@ SECTIONS
             <= CMSE_VENEER_REGION_SIZE, "Veneer region overflowed")
 #endif
 
-    .ER_TFM_CODE : ALIGN(4)
+    .tfm_rodata : ALIGN(4)
     {
-        *(.text*)
-
-        KEEP(*(.init))
-        KEEP(*(.fini))
-
-
-        /* .ctors */
-        *crtbegin.o(.ctors)
-        *crtbegin?.o(.ctors)
-        *(EXCLUDE_FILE(*crtend?.o *crtend.o) .ctors)
-        *(SORT(.ctors.*))
-        *(.ctors)
-
-        /* .dtors */
-         *crtbegin.o(.dtors)
-         *crtbegin?.o(.dtors)
-         *(EXCLUDE_FILE(*crtend?.o *crtend.o) .dtors)
-         *(SORT(.dtors.*))
-         *(.dtors)
-
         *(.rodata*)
 
         KEEP(*(.eh_frame*))
--- a/platform/ext/target/nordic_nrf/nrf9160/partition/region_defs.h
+++ b/platform/ext/target/nordic_nrf/nrf9160/partition/region_defs.h
@@ -104,7 +104,7 @@
 #define S_CODE_LIMIT    (S_CODE_START + S_CODE_SIZE - 1)
 
 #define S_DATA_START    (S_RAM_ALIAS(0x0))
-#define S_DATA_SIZE     (TOTAL_RAM_SIZE / 4) /* 64 KB */
+#define S_DATA_SIZE     ((TOTAL_RAM_SIZE / 2) - 0x4000) /* 112 KB */
 #define S_DATA_LIMIT    (S_DATA_START + S_DATA_SIZE - 1)
 
 /* The CMSE veneers shall be placed in an NSC region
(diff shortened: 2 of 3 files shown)


platform/ext/common/gcc/tfm_common_s.ld
```diff
@@ -602,6 +602,29 @@ SECTIONS
 
 #endif /* TFM_LVL != 1 */
 
+    .ER_TFM_CODE : ALIGN(4)
```
oyvindronningstad, line 605: I tried without the tfm_common.ld commit, and it works fine for me.

oyvindronningstad (commented):
Fantastic work!

Unsure if we need the ld change.

oyvindronningstad:
> In order to use this branch in Zephyr, in place of its [fork of TF-M](https://github.com/zephyrproject-rtos/trusted-firmware-m/tree/master/trusted-firmware-m), some more adaptations from its trusted-firmware-m module need to be applied, see [anangl@c761f2e](https://github.com/anangl/trusted-firmware-m/commit/c761f2efffba3bee6eb4d4a4aa786188d18f5926).

Those are from upstream, right? We could do another upmerge.

platform/ext/common/gcc/tfm_common_s.ld
anangl: I could not build either Regression or RegressionIPC configuration (using gcc 7.3.1) without tweaking the .ld file. Same when building the Zephyr samples (gcc 9.2.0 there), the secure image just did not fit in FLASH without the change.
But since it works for you, perhaps I just need to correct something in my setup.

anangl:
> > In order to use this branch in Zephyr, in place of its [fork of TF-M](https://github.com/zephyrproject-rtos/trusted-firmware-m/tree/master/trusted-firmware-m), some more adaptations from its trusted-firmware-m module need to be applied, see [anangl@c761f2e](https://github.com/anangl/trusted-firmware-m/commit/c761f2efffba3bee6eb4d4a4aa786188d18f5926).
> 
> Those are from upstream, right? We could do another upmerge.

No, those are from the Zephyr's trusted-firmware-m module. This module contains TF-M files copied from tag TF-Mv1.0 (see https://github.com/zephyrproject-rtos/trusted-firmware-m/commit/f966dcb1e4ae9859791797f690bbd53f3aaed6fb) plus those tweaks added in several commits. I just created diff between the tag and TF-M files from this module, and I applied it to my branch.

oyvindronningstad:
> > > In order to use this branch in Zephyr, in place of its [fork of TF-M](https://github.com/zephyrproject-rtos/trusted-firmware-m/tree/master/trusted-firmware-m), some more adaptations from its trusted-firmware-m module need to be applied, see [anangl@c761f2e](https://github.com/anangl/trusted-firmware-m/commit/c761f2efffba3bee6eb4d4a4aa786188d18f5926).
> > 
> > 
> > Those are from upstream, right? We could do another upmerge.
> 
> No, those are from the Zephyr's trusted-firmware-m module. This module contains TF-M files copied from tag TF-Mv1.0 (see [zephyrproject-rtos/trusted-firmware-m@f966dcb](https://github.com/zephyrproject-rtos/trusted-firmware-m/commit/f966dcb1e4ae9859791797f690bbd53f3aaed6fb)) plus those tweaks added in several commits. I just created diff between the tag and TF-M files from this module, and I applied it to my branch.

Ah right. I remember seeing them in review upstream as well:

https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/3922
https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/4335 (merged)
https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/4369 (merged)

platform/ext/common/gcc/tfm_common_s.ld
oyvindronningstad: Maybe you could try once more with the commit completely removed? If you have problems to do with the gcc version we need to accommodate that since there are multiple valid gcc versions.

anangl: I checked once more. Indeed, I can build the Regression configuration without those changes in the linker script (I'm not sure why it didn't work for me earlier), but not the RegressionIPC one. Have you tried to build it?

oyvindronningstad: No, I haven't tried RegressionIPC. We might need the change after all then.

oyvindronningstad (approved):

platform/ext/common/gcc/tfm_common_s.ld
oyvindronningstad: I checked now and ConfigRegressionIPC fails for me as well.

anangl:
Rebased on master, to remove commits merged in #42.

platform/ext/target/nordic_nrf/nrf9160/partition/region_defs.h
```diff
@@ -104,7 +104,7 @@
 #define S_CODE_LIMIT    (S_CODE_START + S_CODE_SIZE - 1)
 
 #define S_DATA_START    (S_RAM_ALIAS(0x0))
-#define S_DATA_SIZE     (TOTAL_RAM_SIZE / 4) /* 64 KB */
+#define S_DATA_SIZE     ((TOTAL_RAM_SIZE / 2) - 0x4000) /* 112 KB */
```
oyvindronningstad, line 107: Can we use defines instead of magic values here?

oyvindronningstad:
I propose we merge this now, and if we don't find a good way to integrate this into the common ld, we create a platform-specific ld before upstreaming.

oyvindronningstad:
> I propose we merge this now, and if we don't find a good way to integrate this into the common ld, we create a platform-specific ld before upstreaming.

@ioannisg Do you agree? Can we merge this? It would streamline the upmerge process.

oyvindronningstad:
> 
> 
> > I propose we merge this now, and if we don't find a good way to integrate this into the common ld, we create a platform-specific ld before upstreaming.
> 
> @ioannisg Do you agree? Can we merge this? It would streamline the upmerge process.

@ioannisg ?

ioannisg (approved):

ioannisg:
This has been merged as part of #43, right @anangl ?
Why does this PR appear as merge-able then?

anangl:
> This has been merged as part of #43, right @anangl ?
> Why does this PR appear as merge-able then?

Perhaps because there are no conflicts (as exactly the same changes were included in #43). It is possible to merge the PR then. But after rebasing it, its commits would become empty, so I guess merging it would be actually a no-op.

PR #43, opened 2020-06-25
Upmerge

oyvindronningstad:
NCS-style upmerge to TFM 1.1-RC1.

Currently contains #41 and ~#45~#46 so that tests can be run.

Not all tests are passing.


oyvindronningstad:
Note that this currently doesn't build because of some renaming. I'm working on updated patches.

ioannisg:
@oyvindronningstad can you provide some details on the failing tests, and possibly the reasons for the failures?

oyvindronningstad:
It seems to be a problem with storage. For Regression, just about all ITS and PS tests fail + a crypto test which loads a key. RegressionIPC fails to start because of a failure in tfm_its_init. It seems to fail while initializing the PS part.

oyvindronningstad:
Can someone else try as well? I tested on nrf53.

ioannisg:
I tried to build the basic configurations - they all build successfully. I 've tested the Default configuration on nRF53 and nrF91 - it works fine. I've also managed to run the IPC configurations , with and without Level2 isolation, and this works, too.

I can successfully build the Regression configurations, though I have not run them on target.

anangl:
@oyvindronningstad You need to rename `SST_CREATE_FLASH_LAYOUT` to `PS_CREATE_FLASH_LAYOUT` (as the Secure Storage service has been renamed to Protected Storage) in 
https://github.com/NordicPlayground/trusted-firmware-m/blob/d3b15f3ebf8dd83866722370b1a1eff74b3b59cb/platform/ext/target/nordic_nrf/common/nrf_common.cmake#L144
All those failing tests should pass then.

oyvindronningstad:
> https://github.com/NordicPlayground/trusted-firmware-m/blob/d3b15f3ebf8dd83866722370b1a1eff74b3b59cb/platform/ext/target/nordic_nrf/common/nrf_common.cmake#L144

Thank you very much! I made the change, and also had to tweak a few values in flash_layout.h. Now Regression and RegressionIPC both pass on nrf53. @anangl @ioannisg please review the changes in the fixup commits.

platform/ext/target/nordic_nrf/common/nrf_common.cmake
```diff
@@ -68,6 +68,8 @@ elseif(BUILD_UART_STDOUT)
     message(FATAL_ERROR "Configuration variable SECURE_UART1 (true|false) is undefined!")
   elseif(SECURE_UART1)
     add_definitions(-DSECURE_UART1)
+  else()
+    message(WARNING "nRF platforms likely need SECURE_UART1 enabled.")
```
anangl, line 72: Currently, configurations without `SECURE_UART1` being set will not work on nRF platforms.
Perhaps we should then force this option to be enabled, like it was so far, and move this line:
https://github.com/NordicPlayground/trusted-firmware-m/blob/4c09829eb9307494c62e58650925af6f33f823c5/platform/ext/target/nordic_nrf/nrf5340/nrf5340.cmake#L20
before that one:
https://github.com/NordicPlayground/trusted-firmware-m/blob/4c09829eb9307494c62e58650925af6f33f823c5/platform/ext/target/nordic_nrf/nrf5340/nrf5340.cmake#L17

Actually, I'm not sure actually why it works in the current form in our current master. I guess it should be anyway corrected.
Same applies to nRF9160.

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
```diff
@@ -177,7 +177,7 @@
 /* Number of PS_SECTOR_SIZE per block */
 #define PS_SECTORS_PER_BLOCK   (0x1)
 /* Specifies the smallest flash programmable unit in bytes */
-#define PS_FLASH_PROGRAM_UNIT  (0x1)
+#define PS_FLASH_PROGRAM_UNIT  4
```
anangl: What about nRF9160? I think we need similar corrections for that platform.

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
```diff
@@ -136,12 +136,12 @@
 
 #define FLASH_PS_AREA_OFFSET            (FLASH_AREA_SCRATCH_OFFSET + \
                                          FLASH_AREA_SCRATCH_SIZE)
-#define FLASH_PS_AREA_SIZE              (0x4000)   /* 16 KB */
+#define FLASH_PS_AREA_SIZE              (0x5000)   /* 20 KB */
```
anangl: The layout description at the beginning of the file should be updated, too.
And nRF9160 should be updated similarly.

But do we need this change, actually? I saw that not all platforms were using these larger sizes (20 kB and 16 kB) of storage areas. Some, like NXP and STM ones, and Musca-A are still using 16/8 kB.

platform/ext/target/nordic_nrf/common/nrf_common.cmake
oyvindronningstad: It seems it was changed in https://github.com/NordicPlayground/trusted-firmware-m/commit/6cbeba6dc10655645d438231d3c23c674f1f4864. We can do an upstream patch to change it to

    if (NOT DEFINED SECURE_UART1)
      set (SECURE_UART1 OFF)
    endif()

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
oyvindronningstad: We have space for it, but I can try it out to see if which exact changes are needed to get the tests to run.

platform/ext/target/nordic_nrf/common/nrf_common.cmake
oyvindronningstad: Added a commit for this.

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
oyvindronningstad: Done

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
oyvindronningstad: It wasn't needed so I reverted these changes.

oyvindronningstad:
I squashed all pre-upmerge commits into a single commit, except for the few commits that touch existing files.

All changes to adapt to changes in upstream TF-M are in fixup commits, and the required commits from #41 and #46 are in separate commits.

I tested this now, and all S/NS tests in 
- Regression
- RegressionIPC, and
- RegressionIPCTfmLevel2

pass on nRF53 and nRF91 :D

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
```diff
@@ -213,7 +212,7 @@
 #define TFM_NV_COUNTERS_AREA_ADDR    FLASH_NV_COUNTERS_AREA_OFFSET
 #define TFM_NV_COUNTERS_AREA_SIZE    (0x18) /* 24 Bytes */
 #define TFM_NV_COUNTERS_SECTOR_ADDR  FLASH_NV_COUNTERS_AREA_OFFSET
-#define TFM_NV_COUNTERS_SECTOR_SIZE  FLASH_AREA_IMAGE_SECTOR_SIZE
+#define TFM_NV_COUNTERS_SECTOR_SIZE  4
```
anangl: This macro does not seem to be used currently (it is only required to be defined) and I think it would be better to leave it as it was defined originally, for consistency with other platforms (it's hard to tell what value would be actually correct here).

anangl (approved):
I tested the above configurations, too. And I can confirm that all tests pass now.

platform/ext/target/nordic_nrf/nrf5340/partition/flash_layout.h
oyvindronningstad: So, I should reset it to FLASH_AREA_IMAGE_SECTOR_SIZE?

anangl: I think so.

oyvindronningstad: Done

oyvindronningstad:
Squashed the fixups. No change to the tree.

ioannisg (approved):

PR #44, opened 2020-07-06
platform: nordic_nrf: Correct names of Non-Secure TIMER handlers

anangl:
Those are supposed to be TIMER1_Handler, not TIMER1_IRQHandler
(see [core_ns_positive_testsuite.c](https://github.com/NordicPlayground/trusted-firmware-m/blob/master/test/suites/core/non_secure/core_ns_positive_testsuite.c#L492)).


ioannisg (approved):

oyvindronningstad (approved):
If it makes it work, go ahead and merge.

It looks like the convention in the other platforms is _Handler rather than _IRQHandler across the board. Though this should be an implementation detail hidden in the platform. So the ideal thing would be to change core_ns_positive_testsuite.c, but it's not a huge deal.

PR #45, opened 2020-07-30
tfm core: make has_access_to_region() compatible with vendor IDAUs

ioannisg:
We enhance the implementation of has_access_to_region(), so
it works with IDAU implementations, i.e. other than the ARM
SAU.

### DIFF
--- a/platform/ext/target/nordic_nrf/common/native_drivers/spu.c
+++ b/platform/ext/target/nordic_nrf/common/native_drivers/spu.c
@@ -21,6 +21,9 @@
 #define FLASH_SECURE_ATTRIBUTION_REGION_SIZE SPU_FLASH_REGION_SIZE
 #define SRAM_SECURE_ATTRIBUTION_REGION_SIZE  SPU_SRAM_REGION_SIZE
 
+#define FLASH_SECURE_ATTRIBUTION_REGIONS_START_ID 0
+#define SRAM_SECURE_ATTRIBUTION_REGIONS_START_ID  64
+
 #define NUM_FLASH_SECURE_ATTRIBUTION_REGIONS \
 	(FLASH_TOTAL_SIZE / FLASH_SECURE_ATTRIBUTION_REGION_SIZE)
 #define NUM_SRAM_SECURE_ATTRIBUTION_REGIONS \
@@ -156,6 +159,66 @@ void spu_regions_flash_config_non_secure_callable(uint32_t start_addr,
 		FLASH_NSC_SIZE_REG(nsc_size),
 		FLASH_NSC_REGION_FROM_ADDR(start_addr),
 		1 /* Lock */);
+}
+
+uint32_t spu_regions_flash_get_base_address_in_region(uint32_t region_id)
+{
+	return FLASH_BASE_ADDRESS +
+		((region_id - FLASH_SECURE_ATTRIBUTION_REGIONS_START_ID) *
+			FLASH_SECURE_ATTRIBUTION_REGION_SIZE);
+}
+
+uint32_t spu_regions_flash_get_last_address_in_region(uint32_t region_id)
+{
+	return FLASH_BASE_ADDRESS +
+		((region_id - FLASH_SECURE_ATTRIBUTION_REGIONS_START_ID + 1) *
+			FLASH_SECURE_ATTRIBUTION_REGION_SIZE) - 1;
+}
+
+uint32_t spu_regions_flash_get_start_id(void) {
+
+	return FLASH_SECURE_ATTRIBUTION_REGIONS_START_ID;
+}
+
+uint32_t spu_regions_flash_get_last_id(void) {
+
+	return FLASH_SECURE_ATTRIBUTION_REGIONS_START_ID +
+		NUM_FLASH_SECURE_ATTRIBUTION_REGIONS - 1;
+}
+
+uint32_t spu_regions_flash_get_region_size(void) {
+
+	return FLASH_SECURE_ATTRIBUTION_REGION_SIZE;
+}
+
+uint32_t spu_regions_sram_get_base_address_in_region(uint32_t region_id)
+{
+	return SRAM_BASE_ADDRESS +
+		((region_id - SRAM_SECURE_ATTRIBUTION_REGIONS_START_ID) *
+			SRAM_SECURE_ATTRIBUTION_REGION_SIZE);
+}
+
+uint32_t spu_regions_sram_get_last_address_in_region(uint32_t region_id)
+{
+	return SRAM_BASE_ADDRESS +
+		((region_id - SRAM_SECURE_ATTRIBUTION_REGIONS_START_ID + 1) *
+			SRAM_SECURE_ATTRIBUTION_REGION_SIZE) - 1;
+}
+
+uint32_t spu_regions_sram_get_start_id(void) {
+
+	return SRAM_SECURE_ATTRIBUTION_REGIONS_START_ID;
+}
+
+uint32_t spu_regions_sram_get_last_id(void) {
+
+	return SRAM_SECURE_ATTRIBUTION_REGIONS_START_ID +
+		NUM_SRAM_SECURE_ATTRIBUTION_REGIONS - 1;
+}
+
+uint32_t spu_regions_sram_get_region_size(void) {
+
+	return SRAM_SECURE_ATTRIBUTION_REGION_SIZE;
 }
 
 void spu_peripheral_config_secure(uint32_t periph_base_addr, bool periph_lock)
--- a/platform/ext/target/nordic_nrf/common/spm_hal.c
+++ b/platform/ext/target/nordic_nrf/common/spm_hal.c
@@ -321,3 +321,68 @@ enum tfm_plat_err_t tfm_spm_hal_nvic_interrupt_enable(void)
 {
     return nvic_interrupt_enable();
 }
+
+bool tfm_spm_hal_has_access_to_region(const void *p, size_t s,
+                                              int flags)
+{
+    cmse_address_info_t tt_base = cmse_TT((void *)p);
+    cmse_address_info_t tt_last = cmse_TT((void *)((uint32_t)p + s - 1));
+
+    uint32_t base_spu_id = tt_base.flags.idau_region;
+    uint32_t last_spu_id = tt_last.flags.idau_region;
+
+    size_t size;
+    uint32_t p_start = (uint32_t)p;
+    int i;
+
+    if ((base_spu_id >= spu_regions_flash_get_start_id()) &&
+        (last_spu_id <= spu_regions_flash_get_last_id())) {
+
+        size = spu_regions_flash_get_last_address_in_region(base_spu_id) + 1 - p_start;
+
+        if (cmse_check_address_range((void *)p_start, size, flags) == 0) {
+            return false;
+        }
+
+        for (i = base_spu_id + 1; i < last_spu_id; i++) {
+            p_start = spu_regions_flash_get_base_address_in_region(i);
+            if (cmse_check_address_range((void *)p_start,
+                spu_regions_flash_get_region_size(), flags) == 0) {
+                return false;
+            }
+        }
+
+        p_start = spu_regions_flash_get_base_address_in_region(last_spu_id);
+        size = (uint32_t)p + s - p_start;
+        if (cmse_check_address_range((void *)p_start, size, flags) == 0) {
+            return false;
+        }
+
+
+    } else if ((base_spu_id >= spu_regions_sram_get_start_id()) &&
+        (last_spu_id <= spu_regions_sram_get_last_id())) {
+
+        size = spu_regions_sram_get_last_address_in_region(base_spu_id) + 1 - p_start;
+        if (cmse_check_address_range((void *)p_start, size, flags) == 0) {
+            return false;
+        }
+
+        for (i = base_spu_id + 1; i < last_spu_id; i++) {
+            p_start = spu_regions_sram_get_base_address_in_region(i);
+            if (cmse_check_address_range((void *)p_start,
+                spu_regions_sram_get_region_size(), flags) == 0) {
+                return false;
+            }
+        }
+
+        p_start = spu_regions_sram_get_base_address_in_region(last_spu_id);
+        size = (uint32_t)p + s - p_start;
+        if (cmse_check_address_range((void *)p_start, size, flags) == 0) {
+            return false;
+        }
+    } else {
+        return false;
+    }
+
+    return true;
+}
--- a/secure_fw/core/tfm_core_mem_check.c
+++ b/secure_fw/core/tfm_core_mem_check.c
@@ -8,6 +8,7 @@
 #include <arm_cmse.h>
 #include "region_defs.h"
 #include "secure_fw/spm/spm_api.h"
+#include "tfm_spm_hal.h"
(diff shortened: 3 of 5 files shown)


secure_fw/core/tfm_core_mem_check.c
```diff
@@ -35,6 +35,28 @@ static enum tfm_status_e has_access_to_region(const void *p, size_t s,
     range_access_allowed_by_mpu =
                           cmse_check_address_range((void *)p, s, flags) != NULL;
 
+    if (!range_access_allowed_by_mpu) {
+    	/*
+    	 * Verification failure may be due to address range crossing
```
oyvindronningstad: Why does it fail if it crosses an IDAU boundary? Does it always fail when crossing an IDAU boundary?

secure_fw/core/tfm_core_mem_check.c
```diff
@@ -35,6 +35,28 @@ static enum tfm_status_e has_access_to_region(const void *p, size_t s,
...
+         * access permissions.
+    	 */
+        const void *p_addr_base = p;
+        const void *p_addr_last = p + s - 1;
+
+        cmse_address_info_t addr_info_base = cmse_TT((void *)p_addr_base);
+        cmse_address_info_t addr_info_last = cmse_TT((void *)p_addr_last);
+
+    	if ((addr_info_base.flags.mpu_region_valid != 0) &&
+            (addr_info_last.flags.mpu_region_valid != 0) &&
```
oyvindronningstad: What does `mpu_region_valid` signify?

secure_fw/core/tfm_core_mem_check.c
```diff
@@ -35,6 +35,28 @@ static enum tfm_status_e has_access_to_region(const void *p, size_t s,
...
+    	 * Verification failure may be due to address range crossing
+         * IDAU boundaries. We check if the whole range is part of a
+         * single MPU region and that the region satisfies the given
+         * access permissions.
+    	 */
+        const void *p_addr_base = p;
+        const void *p_addr_last = p + s - 1;
+
+        cmse_address_info_t addr_info_base = cmse_TT((void *)p_addr_base);
+        cmse_address_info_t addr_info_last = cmse_TT((void *)p_addr_last);
```
oyvindronningstad: Say the mpu region spans 3 IDAU regions, could the middle IDAU region be configured such (e.g. if the middle region is non-secure) that this new code sets `range_access_allowed_by_mpu` to true below when it should have been false?

ioannisg:
@oyvindronningstad I prefer to disregard your feedback here, since i've rewritten most of the patch. New feedback is welcome

platform/ext/target/nordic_nrf/common/spm_hal.c
```diff
@@ -321,3 +321,68 @@ enum tfm_plat_err_t tfm_spm_hal_nvic_interrupt_enable(void)
...
+
+bool tfm_spm_hal_has_access_to_region(const void *p, size_t s,
+                                              int flags)
+{
+    cmse_address_info_t tt_base = cmse_TT((void *)p);
+    cmse_address_info_t tt_last = cmse_TT((void *)((uint32_t)p + s - 1));
+
+    uint32_t base_spu_id = tt_base.flags.idau_region;
+    uint32_t last_spu_id = tt_last.flags.idau_region;
+
```
oyvindronningstad, line 333: To be safe, assert that base_spu_id <= last_spu_id

platform/ext/target/nordic_nrf/common/spm_hal.c
```diff
@@ -321,3 +321,68 @@ enum tfm_plat_err_t tfm_spm_hal_nvic_interrupt_enable(void)
...
+        }
+
+        p_start = spu_regions_sram_get_base_address_in_region(last_spu_id);
+        size = (uint32_t)p + s - p_start;
+        if (cmse_check_address_range((void *)p_start, size, flags) == 0) {
+            return false;
+        }
+    } else {
+        return false;
+    }
```
oyvindronningstad, line 385: Unless I'm missing something, I feel this could be acheived a lot simpler, something like
```suggestion
    for (i = base_spu_id; i <= last_spu_id; i++) {
        uint32_t start = max(spu_get_base_address_in_region(i), p_start);
        uint32_t end = min(spu_get_last_address_in_region(i) + 1, p_start + s);
        if (cmse_check_address_range((void *)start, end - start, flags) == 0) {
            return false;
        }
    }
```
You'd have to merge the `spu_get_base_address_in_region()` etc as I suggested above.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
```diff
@@ -158,6 +161,66 @@ void spu_regions_flash_config_non_secure_callable(uint32_t start_addr,
 		1 /* Lock */);
 }
 
+uint32_t spu_regions_flash_get_base_address_in_region(uint32_t region_id)
+{
+	return FLASH_BASE_ADDRESS +
+		((region_id - FLASH_SECURE_ATTRIBUTION_REGIONS_START_ID) *
+			FLASH_SECURE_ATTRIBUTION_REGION_SIZE);
+}
```
oyvindronningstad, line 169: How about merging flash and ram to something like this?
```suggestion
uint32_t spu_get_base_address_in_region(uint32_t region_id)
{
	if ((base_spu_id >= spu_regions_flash_get_start_id()) &&
		(last_spu_id <= spu_regions_flash_get_last_id())) {
		return FLASH_BASE_ADDRESS +
			((region_id - FLASH_SECURE_ATTRIBUTION_REGIONS_START_ID) *
				FLASH_SECURE_ATTRIBUTION_REGION_SIZE);
	} else if ((base_spu_id >= spu_regions_sram_get_start_id()) &&
		(last_spu_id <= spu_regions_sram_get_last_id())) {	
		return SRAM_BASE_ADDRESS +
			((region_id - SRAM_SECURE_ATTRIBUTION_REGIONS_START_ID) *
				SRAM_SECURE_ATTRIBUTION_REGION_SIZE);
	}
	return 0;
}
```
Likewise for `spu_get_last_address_in_region()` and `spu_get_region_size()`

It would mean you can avoid duplication in `tfm_spm_hal_has_access_to_region()`.

platform/ext/target/nordic_nrf/common/native_drivers/spu.c
```diff
@@ -158,6 +161,66 @@ void spu_regions_flash_config_non_secure_callable(uint32_t start_addr,
...
+		((region_id - SRAM_SECURE_ATTRIBUTION_REGIONS_START_ID + 1) *
+			SRAM_SECURE_ATTRIBUTION_REGION_SIZE) - 1;
+}
+
+uint32_t spu_regions_sram_get_start_id(void) {
+
+	return SRAM_SECURE_ATTRIBUTION_REGIONS_START_ID;
+}
+
+uint32_t spu_regions_sram_get_last_id(void) {
```
oyvindronningstad, line 213: Consider adding a `spu_region_id_is_valid()` which checks that the id is within the FLASH or SRAM region.

oyvindronningstad (commented):
A few initial thoughts.

platform/ext/target/nordic_nrf/common/spm_hal.c
ioannisg: Yes, we can assert, since that should never happen.

platform/ext/target/nordic_nrf/common/spm_hal.c
ioannisg: I like the idea of the min and max (though, actually, they add conditional check, i.e. some overhead)

Also merging the spu_get_.... functions will require conditional checks inside the functions, to distinguish between flash, ram and invalid regions.
Finally, we cannot just enumerate  on -i- because we need to cover the case of base_spu_id pointing to FLASH and last_spu_id pointing to SRAM. Some extra checks will be needed there.

I think the main point in this function would be to be fast, not have minimal footprint. We are targeting nrf53 and 91 with lots of flash. We care more about processing time.

oyvindronningstad: How often will this function be called?

platform/ext/target/nordic_nrf/common/spm_hal.c
```diff
@@ -321,3 +321,68 @@ enum tfm_plat_err_t tfm_spm_hal_nvic_interrupt_enable(void)
...
+    size_t size;
+    uint32_t p_start = (uint32_t)p;
+    int i;
+
+    if ((base_spu_id >= spu_regions_flash_get_start_id()) &&
+        (last_spu_id <= spu_regions_flash_get_last_id())) {
+
+        size = spu_regions_flash_get_last_address_in_region(base_spu_id) + 1 - p_start;
+
+        if (cmse_check_address_range((void *)p_start, size, flags) == 0) {
```
anangl, line 343: Wouldn't it be sufficient to check only one byte in each region (permissions for other bytes within the same region are the same, right)? This would simplify the implementation quite a bit and also `cmse_check_address_range()` would execute faster then (since regions are aligned at 32-byte boundary, when the whole checked address range fits into one 32-byte block, the function executes only one TT instruction).

ioannisg: No, @anangl we really need to check all region(s) because we don't know if the access-check is failing due to crossing IDAU boundaries, or due to violating MPU or SAU configuration. 

By checking within each of the SPU regions we know that any checking failures should be reported to the caller as legitimate.

anangl: Right, I haven't thought about MPU in this context.

secure_fw/core/tfm_core_mem_check.c
```diff
@@ -35,6 +36,23 @@ static enum tfm_status_e has_access_to_region(const void *p, size_t s,
...
+    if (!range_access_allowed_by_mpu) {
+        /*
+         * Verification failure may be due to address range crossing
+         * one or multiple IDAU boundaries. In this case request a
+         * platform-specific check for access permissions.
+         */
+        cmse_address_info_t addr_info_base = cmse_TT((void *)p);
+        cmse_address_info_t addr_info_last = cmse_TT((void *)((uint32_t)p + s - 1));
+
+        if ((addr_info_base.flags.idau_region_valid != 0) &&
```
anangl, line 49: I guess you'll add some `#if` here, to not force all platforms to provide an implementation of `tfm_spm_hal_has_access_to_region()`.

ioannisg: Well, @anangl this is a good point. I don't know what the tf-m upstream project favors here. 
One option is what you suggest - an #ifdef IDAU_PRESENT, or sthg like this. 

The other option is to keep the check for _all_ platforms, regardless of IDAU existence. If the IDAU does not exist, then the idau_region_valid bits are zero, and we do not enter the if. For likning purposes, of course, we need to provide a __weak empty implementation of this returning false. This is only for linking: this function won't ever enter.
What do you think?

Also a question to @oyvindronningstad .

anangl: > The other option is to keep the check for _all_ platforms, regardless of IDAU existence. If the IDAU does not exist, then the idau_region_valid bits are zero, and we do not enter the if.

Yes, this will be a dead code. But the compiler won't be able to detect it. To me it feels natural that there should be a possibility to get rid of this part, so I think some `#if` is the way to go.

ioannisg: then @anangl would you like something such as IDAU_PRESENT? This complies with ARM CMSIS macros like MPU_PRESENT, FPU_PRESENT.

There is another option: use `#ifdef SAU_REGION_PRESENT == 0`.
SAU_REGION_PRESENT is defined in the SoC CMSIS headers. In nRF MDK this is defined as zero, indicating that there is no SAU and, implicitly, there is an implementation-defined IDAU. I guess if SAU exists, there is no IDAU ( cannot imagine a SoC that has both).

anangl: Yes, that option with SAU regions present symbol sounds reasonable. However, it looks like it should be `__SAUREGION_PRESENT`, not `__SAU_REGION_PRESENT` like it is currently defined in MDK. :)

ioannisg: Good point. We should use the right CMSIS header then. Anyway, for nRF this should be _not_ defined or defined as zero

ioannisg: Updated @anangl , thanks.

platform/ext/target/nordic_nrf/common/spm_hal.c
```diff
@@ -321,3 +321,68 @@ enum tfm_plat_err_t tfm_spm_hal_nvic_interrupt_enable(void)
 {
     return nvic_interrupt_enable();
 }
+
+bool tfm_spm_hal_has_access_to_region(const void *p, size_t s,
+                                              int flags)
+{
```
anangl, line 327: Did you consider moving the implementation of this function entirely to `spu.c` (and just calling it from here)? This would reduce the code footprint and execution time by eliminating the need of several `spu_regions_*` calls.
I am wondering if it will be actually beneficial to have this whole `spu_regions_*` API. Do you foresee some other use cases for it?

ioannisg: Thing is, I recall, we wanted the spu.c to be, eventually, an nRFX driver for the system protection unit exposing a relevant API, and should not contain CMSE checks that are tight to the ARM Core. That's why I favor this separation, in the first place.

If we decide to move spu. to nrfx, we cannot have functions that use the TT intrinsics, or other v8-m or CMSE functionality.

ioannisg: (Also, strictly, speaking, we cannot call the spu.c API from tfm-core; there _has_ to do a platform wrapper function, at least a transparent wrapper.)

anangl: > If we decide to move spu. to nrfx, we cannot have functions that use the TT intrinsics, or other v8-m or CMSE functionality.

I don't see why we cannot, actually. We already use some ARM specific things (like barriers) in drivers that need it.

> (Also, strictly, speaking, we cannot call the spu.c API from tfm-core; there _has_ to do a platform wrapper function, at least a transparent wrapper.)

I was thinking about a wrapper. That's why I wrote "(and just calling it from here)".

ioannisg: In my opinion, the implementation should not part of the SPU driver. I don't want to have a function inside the SPU driver that assesses memory accesses, _not_ explicitly based on SPU, but based on MPU as well, and based on other TrustZone configuration, e..g the SAU ALL_NS bit flag. The function is beyond the responsibility of the SPU peripheral, and, therfore, it has to be outside the driver, in some common file, like the spm_hal.

anangl (approved):
I'm okay with how it is now. It solves the problem it is supposed to. Any optimizations can be done later, if needed.

oyvindronningstad (approved):

ioannisg:
@oyvindronningstad i wonder if this patch is able to solve all the problems we have in the regression test suite, or not. Hope you could do some testing :)

oyvindronningstad:
Together with #41, this makes both Regression and RegressionIPC pass fully.

oyvindronningstad:
I guess we can merge?

PR #46, opened 2020-08-13
Fix has access to region draft

ioannisg:


ioannisg:
This has been merged as part of #43

PR #48, opened 2020-09-07
PSA API tests

oyvindronningstad:
Fixes needed for PSA API tests, for nrf53 and nrf91. Crypto tests don't work because we need CC310/312. but storage and attestation tests pass (on nrf53).

Porting can be found on https://github.com/oyvindronningstad/psa-arch-tests/pull/1.

To run, first compile the tests (in psa-arch-tests/api-tests/BUILD, so TF-M automatically finds them):

    cmake ../ -GNinja -DCPU_ARCH=armv8m_ml -D"SUITE=STORAGE" -DPSA_INCLUDE_PATHS="/home/oyro/git/ncs/modules/tee/tfm_dev/trusted-firmware-m/interface/include/" -DTARGET=tgt_dev_apis_tfm_nrf
    ninja

Then, compile TF-M:

    cmake .. -GNinja -DCOMPILER=GNUARM -DTARGET_PLATFORM=nrf5340pdk_nrf5340_cpuapp -DPROJ_CONFIG=`readlink -f ../configs/ConfigPsaApiTest.cmake` -DPSA_API_TEST_STORAGE=ON
    ninja install

Then flash. The test output comes on the NS UART (UART0).


ioannisg:
@anangl pls, take a look :)

platform/ext/common/gcc/tfm_common_s.ld
```diff
@@ -1412,6 +1358,36 @@ SECTIONS
     } > RAM AT> FLASH
 #endif
 
+#ifndef TFM_MULTI_CORE_TOPOLOGY
+    /*
+     * Place the CMSE Veneers (containing the SG instructions) in a correctly
+     * aligned region so that the SAU can programmed to just set this region as
```
ioannisg: SAU or SPU?

platform/ext/target/nordic_nrf/common/nrf_common.cmake
```diff
@@ -18,7 +18,7 @@ if(COMPILER STREQUAL "ARMCLANG")
       set (RTX_LIB_PATH "${CMSIS_DIR}/RTOS2/RTX/Library/ARM/RTX_V8MMN.lib")
     endif()
 elseif(COMPILER STREQUAL "GNUARM")
-    set (S_SCATTER_FILE_NAME   "${PLATFORM_DIR}/common/gcc/tfm_common_s.ld")
+    set (S_SCATTER_FILE_NAME   "${NRF_COMMON_DIR}/gcc/nordic_nrf_s.ld")
```
ioannisg: @oyvindronningstad does this let as completely by pass the common linker script?

oyvindronningstad: Yes

ioannisg: Thanks. So basically, you have reverted all changed to the tfm_common.? @oyvindronningstad ?

platform/ext/target/nordic_nrf/common/nrf_common.cmake
```diff
@@ -147,3 +147,7 @@ elseif(BUILD_FLASH)
   set(ITS_CREATE_FLASH_LAYOUT ON)
   embedded_include_directories(PATH "${NRF_COMMON_DIR}/cmsis_drivers" ABSOLUTE)
 endif()
+
+if(PSA_API_TEST_NS)
+  embedded_set_target_link_defines(TARGET tfm_s DEFINES "PSA_API_TEST_NS")
```
ioannisg: Could you please add a comment here explaining why this is needeD?

ioannisg (commented):
@oyvindronningstad looking at the second commit; why is this requirement for the scatch area not present in the other platform targets?

platform/ext/common/gcc/tfm_common_s.ld
oyvindronningstad: This is the common file. I can change it to "SAU/IDAU".
```suggestion
     * aligned region so that the SAU/IDAU can programmed to just set this region as
```

platform/ext/target/nordic_nrf/common/nrf_common.cmake
oyvindronningstad: Not all of them. There were some general improvements there, which I want to upstream.

platform/ext/target/nordic_nrf/common/nrf_common.cmake
oyvindronningstad:
```suggestion
  # We want to place the veneers in different locations depending on this define.
  embedded_set_target_link_defines(TARGET tfm_s DEFINES "PSA_API_TEST_NS")
```

oyvindronningstad:
> 
> 
> @oyvindronningstad looking at the second commit; why is this requirement for the scatch area not present in the other platform targets?

That's a good question. I'll take a look.

oyvindronningstad:
I refactored things quite a bit now. I tried to minimize the psa-arch-tests port, and instead call into TF-M. I extended plat_test.c to implement som functions needed by the psa-arch-tests port.

As a result, the psa-arch-tests port is agnostic to the specific chip, so I just merged them, and renamed it to tgt_dev_apis_tfm_nrf. Also, you don't need as many include paths when running Cmake. I updated the example in the description.

oyvindronningstad:
cc @andvib @magnev

platform/ext/target/nordic_nrf/common/nrf_common.cmake
```diff
@@ -95,7 +95,7 @@ endif()
 
 if (NOT DEFINED BUILD_PLAT_TEST)
   message(FATAL_ERROR "Configuration variable BUILD_PLAT_TEST (true|false) is undefined!")
-elseif(BUILD_PLAT_TEST)
+elseif(BUILD_PLAT_TEST OR PSA_API_TEST_NS)
```
ioannisg: Instead of doing this @oyvindronningstad I wonder if PSA_API_TEST_NS should just "enable" BUILD_PLAT_TEST

PR #49, opened 2020-09-09
platform: nrf9160: Use 88 kB of RAM for SPE and all the rest for NSPE

anangl:
Use for SPE the minimum amount of RAM (aligned to the SPU region
boundary) that is needed for the RegressionIPC configuration.
All the rest assign to NSPE (so that whole this area is configured
in SPU as non-secure).

### DIFF
--- a/platform/ext/target/nordic_nrf/nrf9160/partition/region_defs.h
+++ b/platform/ext/target/nordic_nrf/nrf9160/partition/region_defs.h
@@ -104,7 +104,10 @@
 #define S_CODE_LIMIT    (S_CODE_START + S_CODE_SIZE - 1)
 
 #define S_DATA_START    (S_RAM_ALIAS(0x0))
-#define S_DATA_SIZE     ((TOTAL_RAM_SIZE / 2) - 0x4000) /* 112 KB */
+/* Assign to SPE the minimum amount of RAM (aligned to the SPU region boundary)
+ * that is needed for the most demanding configuration, which turns out to be
+ * the RegressionIPC one. */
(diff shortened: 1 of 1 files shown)


anangl:
It turns out that after the upmerge, significantly less RAM is needed for the RegressionIPC configuration.

platform/ext/target/nordic_nrf/nrf9160/partition/region_defs.h
```diff
@@ -104,7 +104,10 @@
 #define S_CODE_LIMIT    (S_CODE_START + S_CODE_SIZE - 1)
 
 #define S_DATA_START    (S_RAM_ALIAS(0x0))
-#define S_DATA_SIZE     ((TOTAL_RAM_SIZE / 2) - 0x4000) /* 112 KB */
+/* Assign to SPE the minimum amount of RAM (aligned to the SPU region boundary)
+ * that is needed for the most demanding configuration, which turns out to be
+ * the RegressionIPC one. */
```
oyvindronningstad, line 109: Did you test this on the PSA tests?

oyvindronningstad (approved):

platform/ext/target/nordic_nrf/nrf9160/partition/region_defs.h
anangl: Yes, I tried it with storage tests, and those needed ~62 kB (in the IPC mode, ~48 kB otherwise) of RAM for PSE.

ioannisg (approved):

PR #50, opened 2020-09-10
Add nordic specific secure_fw linker script.

oyvindronningstad:
These are the linker changes from #48.

This leaves tfm_common_s.ld unchanged in our branch.


platform/ext/target/nordic_nrf/common/gcc/nordic_nrf_s.ld.template
```diff
@@ -0,0 +1,691 @@
+;/*
```
anangl, line 1: Do we actually need this file?

oyvindronningstad: Not sure..

oyvindronningstad: Let's keep it until we're sure?

anangl: It does not seem to be used anywhere.

anangl: `tfm_common_s.ld.template` is referenced here: https://github.com/NordicPlayground/trusted-firmware-m/blob/2ac4b29ad8a8f2d5697b1ee3b5157077402690d2/tools/tfm_generated_file_list.yaml#L39-L40

oyvindronningstad: It can be used with a script to regenerate the linker script. Maybe it's needed if we add custom secure services. I haven't fully understood when it's used.

anangl: ok, so let's leave it

oyvindronningstad: Maybe we should add our file to that list? We can do it later if so.

ioannisg (commented):
@oyvindronningstad the second commit is a mirror of the upstream change-request you sent , right?

platform/ext/target/nordic_nrf/common/gcc/nordic_nrf_s.ld.template
```diff
@@ -0,0 +1,691 @@
...
+        Load$$LR$$LR_VENEER$$Base = ABSOLUTE(.);
+        *(.gnu.sgstubs*)
+#ifdef CMSE_VENEER_REGION_IN_CODE
+    } > FLASH
+#else
+    } > VENEERS
+#endif
+
+    /*
+     * This is needed because the linker seems to place the veneers *after* the
```
ioannisg: talking about _this_ part here

ioannisg (commented):
@oyvindronningstad @anangl I am also wondering whether our own linker script should contain the parts under #ifdef TFM_MULTI_CORE_TOPOLOGY.
We won't support this any time soon, right?

platform/ext/target/nordic_nrf/common/gcc/nordic_nrf_s.ld.template
oyvindronningstad: It's sort of the same change, but here it is used in our linker script, but the upstream PR is for the tfm_common_s.ld.

ioannisg: yes of course.

ioannisg (approved):

platform/ext/target/nordic_nrf/common/gcc/nordic_nrf_s.ld.template
anangl: Yes, let's keep this file for now and add it to that list later, if needed (or drop, otherwise). But it seems that the contents of this file should be updated (it doesn't have the veneers stuff defined as a macro).

oyvindronningstad: Thanks, fixed.

anangl (commented):
These entries are still missing in both `nordic_nrf_s.ld*` files:
https://github.com/NordicPlayground/trusted-firmware-m/blob/2ac4b29ad8a8f2d5697b1ee3b5157077402690d2/platform/ext/common/gcc/tfm_common_s.ld#L307-L308

oyvindronningstad:
> 
> 
> These entries are still missing in both `tfm_common_s.ld*` files:
> 
> https://github.com/NordicPlayground/trusted-firmware-m/blob/2ac4b29ad8a8f2d5697b1ee3b5157077402690d2/platform/ext/common/gcc/tfm_common_s.ld#L307-L308

They are removed in the current upstream master.

anangl (approved):

PR #51, opened 2020-09-10
nordic_common.cmake: Remove Cmake warnings that are no longer relevant

oyvindronningstad:
Everything has been implemented now.


anangl (approved):

ioannisg:
Why did we have this thing with the MPC in the first place @oyvindronningstad `?
We would never going to implement it, right?

ioannisg (commented):
@oyvindronningstad I am trying this on nRF53 and still get a cmake warning about RETARGET not implemented. Is this sthg we could also get rid off?

ioannisg (approved):

